PR #16, opened 2022-01-07
added information about accumulators and expanded UDF examples

a-m-thomas:


modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -51,13 +54,10 @@ inline string reverse(string str){
 
 [CAUTION]
 ====
-If any code in `ExprFunctions.hpp` or `ExprUtil.hpp` causes a compilation error, GSQL cannot install any GSQL query, even if the GSQL query doesn't call any query UDF. Therefore, please test each new query UDF after adding it. One way of testing a function is to create a new file `test.cpp` and compile it: +
-`> g{pp} test.cpp
+If any code in `ExprFunctions.hpp` or `ExprUtil.hpp` causes a compilation error, GSQL will be unable to install _any_ new queries, whether user-defined or not. Therefore, please test each new query UDF after adding it. One way of testing a function is to make a new file `test.cpp` and compile it.
```
lennessyy, line 57: This is a little tricky now. Since users can use accumulators in their UDF, those accumulator classes rely on the accumulator definition in GSQL. If they just create a `test.cpp` and put their functions in it, and those functions happen to contain an accumulator, then it actually won't compile. 

I can't think of another quick way of testing it - so perhaps we'll refrain from making a suggestion, but just keep it as a warning.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -51,13 +54,10 @@ inline string reverse(string str){
...
 
-____
-./a.out``
-You might need to remove the include header ``#include <gle/engine/cpplib/headers.hpp>`` in ``ExprFunctions.hpp`` and ``ExprUtil.hpp` in order to compile.
-____
+You might need to remove the included header `#include <gle/engine/cpplib/headers.hpp>` in `ExprFunctions.hpp` and `ExprUtil.hpp` in order to compile.
+====
```
lennessyy, line 60: I think this `#include` header is what allows users to use accumulators. In the current version, it no longer exists in the UDF file you download. When you upload it, the include directive is applied to your UDF code automatically.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -72,22 +72,19 @@ int main () {
...
+[source,gsql]
 ----
 GSQL > PUT ExprFunctions FROM "/path/to/udf_file.hpp"
 PUT ExprFunctions successfully.
 GSQL > PUT ExprUtil FROM "/path/to/utils_file.hpp"
 PUT ExprUtil successfully.
 ----
 
-The `PUT` command will automatically upload the files to all nodes in a cluster. Once the files are uploaded, you will be able to call the query UDF the next time GSQL is executed. This includes the next time you start the GSQL shell or executing GSQL scripts from a bash shell.
+The `PUT` command will automatically upload the files to all nodes in a cluster, overwriting any existing files that contain UDFs. Once the files are uploaded, you will be able to call the query UDF the next time GSQL is executed. This includes the next time you start the GSQL shell or executing GSQL scripts from a bash shell.
```
lennessyy, line 87: Should mention GraphStudio here also. Does user need to refresh the browser?

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -160,20 +154,30 @@ namespace UDIMPL {
 
   // other UDFs ...
```
lennessyy, line 155: Did you see all of this in the UDF file you downloaded? What version of TigerGraph were you using?

lennessyy (changes requested):

lennessyy:
Hi @a-m-thomas , can you resolve the other two comments here also?

lennessyy (changes requested):
The sentence above the table says you can only use `bool`, `int`, `float`, `double`, and `string` (NOT `std::string`), but the table lists other data types. 

Did you forget to change the sentence above the table?

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -158,22 +166,30 @@ namespace UDIMPL {
...
+[source,gsql]
 ----
-GSQL > PUT ExprFunction FROM "/home/tigergraph/documents/udf.hpp"
+GSQL > PUT ExprFunction FROM "/path/to/udf.hpp"
 PUT ExprFunctions successfully.
 ----
 
-The UDF has now been added to GSQL and you can start using the function in GSQL queries.
+The UDF has now been added to GSQL. You can `INSTALL` and `RUN` the function in GSQL queries.
+
```
lennessyy, line 181: 1. A UDF is different from a query. `INSTALL QUERY` and `RUN QUERY` work on queries only. 
2. Don't use code words as English words.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -51,13 +70,8 @@ inline string reverse(string str){
...
 [CAUTION]
 ====
-If any code in `ExprFunctions.hpp` or `ExprUtil.hpp` causes a compilation error, GSQL cannot install any GSQL query, even if the GSQL query doesn't call any query UDF. Therefore, please test each new query UDF after adding it. One way of testing a function is to create a new file `test.cpp` and compile it: +
-`> g{pp} test.cpp
-
-____
-./a.out``
-You might need to remove the include header ``#include <gle/engine/cpplib/headers.hpp>`` in ``ExprFunctions.hpp`` and ``ExprUtil.hpp` in order to compile.
-____
+If any code in `ExprFunctions.hpp` or `ExprUtil.hpp` causes a compilation error, GSQL will be unable to install _any_ new queries, whether user-defined or not.
```
lennessyy, line 73: Don't confuse queries with UDFs.

lennessyy (changes requested):

a-m-thomas:
I removed the include statements and any references to them after testing that the provided ExprFunction does work as shown with only that one function included. I also chose to include a full sequence of commands for installing a test query to demonstrate rng() in case a future user runs into the same confusion I had about which commands apply to queries and functions.

lennessyy (approved):

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -1,46 +1,67 @@
...
+Query UDFs are user-editable in the form of an `.hpp` file locally on your machine. These get uploaded and interpreted by the server through a GSQL command. Creating, editing, or removing Query UDFs is done by downloading them from the server, editing the file, and reuploading.
 
 == Define a query UDF
 
 Below are the steps to add a Query UDF to GSQL:
 
 === Step 1: Download current query UDF file
 
-Use the `GET ExprFunctions` command in GSQL to download the current UDF file to any location on your machine. The file and the directores will be created if they do not exist, and the file must end with the file extention `.hpp`:
+Use the `GET ExprFunctions` command in GSQL to download the current list of functions into a local file. The path can be absolute or relative to your current directory, but the file extension must be `.hpp`:
```
victor-gsl, line 16: "to download the current list of functions into a local file": I'm not sure this is correct either. The command does not download a list.  It downloads a file which contains a _set_ of UDFs.  "If no UDFs have been defined yet, then the file will be empty." <-- Is this true?

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -1,46 +1,67 @@
...
+| `std::string` | *No* | *No* | Yes
+| xref:accumulators.adoc[Accumulators] | Yes | Yes | Yes
+
+| All other C{pp} data types | *No* | *No* | Yes
+|===
 
 If the function requires a user-defined struct or helper function, define it in the `ExprUtil` file you downloaded in Step 1.
 
-Below is an example of a query UDF definition:
+Below is an example of a query UDF definition that includes two separate functions. The first function returns `true` if the passed value is greater than 3, and the second reverses a string.
```
victor-gsl, line 54: "Below is an example of a query UDF definition that includes two separate functions". I believe this is a misuse of _UDF_.  That file contains two UDFs, not one. They aren't even related. It just happens that the user wants to add them at the same time. UDF literally means User-Defined Function, so it makes no sense to say, "This user-defined function contains two functions" (unless they are subfunctions).

victor-gsl (commented):
Comments about clear usage of the terms _UDF_, _function_, and _file_.
For function and file, we should be just sticking with the normal definitions. Note that there are several files in play, so be careful and accurate when referring to them. There are the internal files ExprFunctions.hpp and ExprUtil.hpp, and there are the files that the local files that the user is editing.

_UDF_ should refer to one C++ function that the user has added or wishes to add to the internal ExprFunctions.hpp file. The file can contain multiple UDFs; the user can edit multiple UDFs at a time. The grammar of _UDF_ is exactly the same as that of _function._

PR #27, opened 2022-02-03
V1v2 comparision doc 1026

lennessyy:
Hi @adil-ainihaer , we also want to publish a document that highlights the differences between V1 and V2. I cleaned up our old note and made it into this new page. 

I put every difference that I know of into this document. Please take a look and let me know if I missed anything or misunderstood anything.

### DIFF
--- a/modules/querying/pages/syntax-versions.adoc
+++ b/modules/querying/pages/syntax-versions.adoc
@@ -30,11 +30,11 @@ The path pattern may consist of multiple hops and allows for powerful xref:tutor
 |Edge notation
 |Edges in a step can optionally have a right arrowhead outside the parentheses.
 For example:
-`Start:s -( Edges:e )-> Target:t`
+`Start:s -( Edges:e )-> Target:t` or `Start:s -( Edges:e )- Target:t`
 |No arrowheads are allowed outside the parentheses.
-Each edge may have arrowheads inside the parentheses to show the direction.
+If an edge type is directed, the edge notation must have arrowheads inside the parentheses to indicate direction.
 For example:
-`Start:s -(Edge1>:e1)- Middle:m -(<Edge2:e2)- Target:t`
+`Person:s-(STUDY_AT>:e)-University:t`
 
 
 |Traversal direction
--- a/modules/querying/pages/syntax-versions.adoc
+++ b/modules/querying/pages/syntax-versions.adoc
@@ -74,7 +74,7 @@ This section provides you with instructions to convert a V1 query into a V2 quer
 
 GSQL can detect the syntax version of the query automatically, so there is no need to modify your queries from V1 to V2.
 
-. Delete or arrowheads outside parentheses in the `FROM` clause.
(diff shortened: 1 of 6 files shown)


modules/querying/pages/syntax-versions.adoc
```diff
@@ -0,0 +1,84 @@
...
+`Start:s -(Edge1>:e1)- Middle:m -(<Edge2:e2)- Target:t`
+
+
+|Traversal direction
+|The traversal action is from left to right, and the notation uses either a rightward facing arrowhead or no arrowhead to depict a connection.
+|The traversal direction is under the control of the query writer, with arrowheads on each edge set to show the direction.
+
+|Source vertex set
+|The source vertex set in a step in the `FROM` clause must be xref:declaration-and-assignment-statements.adoc#_vertex_set_variables[a vertex set (seed set)] declared earlier in the query.
+|The xref:select-statement/index.adoc#_source_vertex_set[source vertex set path pattern] in `FROM` clause can be a vertex set, a vertex type, or an accumulator.
```
adil-ainihaer, line 46: The use of accumulators for source is not allowed yet.

modules/querying/pages/syntax-versions.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+|===
+
+== Changing a query from V1 to V2
+GSQL can detect the syntax version of the query automatically, so there is no need to modify your queries from V1 to V2.
+However,
+
+. Delete arrowheads outside parentheses in the `FROM` clause.
+* This is the most common incompatibility issue between a query written in V1 and V2.
+Most V1 queries can run smoothly in V2 after the arrowheads are deleted.
```
adil-ainihaer, line 78: "Most V1 queries can run smoothly in V2 after the arrowheads are deleted."

This is not necessarily true since the edge type can be directed. The more accurate way to say is "... after the arrowheads are deleted to moved."

lennessyy: Added note about arrowheads

modules/querying/pages/syntax-versions.adoc
```diff
@@ -70,11 +70,20 @@ This can be done explicitly or implicitly.
...
 
 == Changing a query from V1 to V2
+This section provides you with instructions to convert a V1 query into a V2 query.
+
 GSQL can detect the syntax version of the query automatically, so there is no need to modify your queries from V1 to V2.
-However,
 
-. Delete arrowheads outside parentheses in the `FROM` clause.
-* This is the most common incompatibility issue between a query written in V1 and V2.
+. Delete or arrowheads outside parentheses in the `FROM` clause.
```
adil-ainihaer, line 77: "Delete or" --> "Delete or move"?

adil-ainihaer (approved):
Looks good overall. Noticed one potential typo issue.

modules/querying/pages/syntax-versions.adoc
lennessyy: Thanks for catching that - fixed

PR #31, opened 2022-02-23
added github hosting instructions

a-m-thomas:


modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -176,3 +176,37 @@ GSQL > RUN QUERY rngExample()
...
+gadmin config set GSQL.GithubPath sample_code/src
+gadmin config apply
+----
+
+The file names must be `ExprFunctions.hpp` and `ExprUtil.hpp`. This is in contrast to the `PUT` method, where the files could have any file name.
+
+[NOTE]
+When the files are hosted on GitHub, the `typedef` statements and `include` directives must be part of the `.hpp` files. These statements are found at `AppRoot/dev/gdk/gsql/src/QueryUdf/ExprFunctions.hpp`.
+
+For more information about the possibilities that GitHub file hosting brings, such as High Availability (HA), please see the corresponding xref:tigergraph-server:ha:ha-for-gsql-server.adoc[High Availability documentation.]
```
lennessyy, line 212: Using Github to store UDFs does not bring HA. 
It is HA that necessitates the use of PUT and GitHub. When we first introduced HA, if a user wanted to change the UDF file, they needed to manually edit the file on every machine. Using PUT or GitHub allows them to edit only once. 

Therefore, there is no need to refer to the HA page.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -176,3 +176,37 @@ GSQL > RUN QUERY rngExample()
...
+[source]
+----
+GSQL.GithubUserAcessToken # the credential, or "anonymous", empty means not using github
+GSQL.GithubRepository     # e.g. tigergraph/ecosys
+GSQL.GithubBranch         # optional, o/w use "master" branch, e.g. demo_github
+GSQL.GithubPath           # path to the directory in the github that has TokenBank.cpp, ExprFunctions.hpp, ExprUtil.hpp, e.g. sample_code/src
+GSQL.GithubUrl            # optional, used for github enterprise, e.g. https://api.github.com
+----
+
+Run the following `gadmin config set` commands to give GSQL access to GitHub.
```
lennessyy, line 196: The commands that follow are only an example configuration. It can be a little misleading to tell users to just run these commands as they might be interpreting it as running the commands as they are, especially if you haven't thoroughly explained what each parameter does. 

A better phrasing would be:

Use the `gadmin config set` (link to gadmin config here) command to configure the value of the aforementioned parameters to connect GSQL to your GitHub repository hosting your UDF file. 
Below is an example configuration.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -176,3 +176,37 @@ GSQL > RUN QUERY rngExample()
...
+gadmin config set GSQL.GithubRepository tigergraph/ecosys
+gadmin config set GSQL.GithubBranch demo_github
+gadmin config set GSQL.GithubPath sample_code/src
+gadmin config apply
+----
+
+The file names must be `ExprFunctions.hpp` and `ExprUtil.hpp`. This is in contrast to the `PUT` method, where the files could have any file name.
+
+[NOTE]
+When the files are hosted on GitHub, the `typedef` statements and `include` directives must be part of the `.hpp` files. These statements are found at `AppRoot/dev/gdk/gsql/src/QueryUdf/ExprFunctions.hpp`.
```
lennessyy, line 210: We should just include the files here with an example.

a-m-thomas: Do you mean to write out the whole source of the files, or host them as downloadable? If the user copies them from a new TigerGraph installation to GitHub, they'd come with the directives intact. That was easier for me to do when testing it out. See what you think of the way I rearranged the instructions.

a-m-thomas:
I need to go through this again for changes since it's still a bit sloppy, but I rearranged into the structure I think is best. I need to run more tests to see how the priority works for GitHub UDFs vs PUT UDFs, and what steps need to be done for when you try to PUT a file with functions that doesn't exist in GitHub.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -51,36 +49,104 @@ Below is a list of data types and where each is allowed in a function.
...
   std::reverse(str.begin(), str.end());
   return str;
 }
 ----
 
+== Use GitHub to store UDFs
+
+You can configure GSQL to read from a GitHub repository for ExprFunctions and ExprUtil.
+
+If GitHub access is configured, GSQL will retrieve user source code files from GitHub before files uploaded via `PUT`, so long as the files exist.
```
lennessyy, line 73: @a-m-thomas Is that the files need to exist or is it that only the configuration needs to be valid?

lennessyy (commented):
Hi Alex, it looks like you still have unresolved conflicts on line 20 and line 79

lennessyy (approved):

PR #34, opened 2022-03-16
add pattern repetition to select statement page

lennessyy:


lennessyy:
Hi @adil-ainihaer , I'm adding the star pattern to the language reference. Can you take a look at the changes I've made and let me know if they are accurate?

adil-ainihaer (approved):

PR #36, opened 2022-03-24
add github instructions and format consistently with query UDFs

a-m-thomas:


modules/ddl-and-loading/pages/add-token-function.adoc
```diff
@@ -47,50 +33,52 @@ below depending on the attribute type.
...
+in GSQL the token function will return the string stored in `oToken`:
 
-* `+oToken+` is the returned string value
-* `+oTokenLen+` is the length of the return string
+* `oToken` is the returned string value
+* `oTokenLen` is the length of the return string
 
-Note that the input tokens are always in string (`+char*+`) format. If
+Note that the input tokens are always in string (`char*`) format. If
 necessary, convert them to other types inside the function.
```
lennessyy, line 73: Did we add a note to say that users must make sure that the tokenbank functions compile, otherwise it impacts all loading jobs? I think that was the original suggestion from support.

lennessyy (approved):

PR #41, opened 2022-04-14
added an example for loading job referencing column names

lennessyy:
- Added an example for load statement referencing column names

### DIFF
--- a/modules/ddl-and-loading/partials/load-statement.adoc
+++ b/modules/ddl-and-loading/partials/load-statement.adoc
@@ -61,7 +61,7 @@ By specifying the vertex type, this tells the loader what id types to expect. Th
 To support fast, out-of-order loading, *if one or both of the endpoint vertices of an edge do not yet exist, the loader will create vertices with the necessary IDs and default attribute values.*
 If the vertex data is loaded later, it will be automatically merged with the automatically created vertices.
 
-The user can disable this feature and perform regular referential integrity checking by using the `VERTEX_MUST_EXIST=true` option.
+The user can disable this feature and perform regular referential integrity checking by setting the `VERTEX_MUST_EXIST` option to `true`.
 ====
 
 === Examples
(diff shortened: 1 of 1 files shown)


modules/ddl-and-loading/partials/load-statement.adoc
```diff
@@ -55,24 +55,13 @@ For the `TO VERTEX` and `TO EDGE` destination clauses, the following rules for i
...
 ====
-*For fast loading of edge data, referential integrity checking is disabled by default.* +
-For an edge to be valid, it must refer to endpoint vertices that exist.
-To support fast, out-of-order loading, *if one or both of the endpoint vertices do not yet exist, the loader will create vertices with the necessary IDs and default attribute values.*
-If the vertex data is loaded later, it will be automatically merged with the dummy vertices.
-The user can disable this feature and perform regular referential integrity checking by setting the `VERTEX_MUST_EXIST=true` option.
+To support fast, out-of-order loading, *if one or both of the endpoint vertices of an edge do not yet exist, the loader will create vertices with the necessary IDs and default attribute values.*
+If the vertex data is loaded later, it will be automatically merged with the automatically created vertices.
+
+The user can disable this feature and perform regular referential integrity checking by using the `VERTEX_MUST_EXIST=true` option.
```
alexthomasTG: "by setting the `VERTEX_MUST_EXIST` option to `true`"

modules/ddl-and-loading/partials/load-statement.adoc
```diff
@@ -128,7 +140,9 @@ Arithmetic Operators:: `+`, `-`, `*`, `/`, `{caret}`. Numeric operators can be u
 
 [WARNING]
 ====
-Because computers necessarily can only store approximations for most `DOUBLE` and `FLOAT` type values, it is not recommended to test these data types for exact equality or inequality. Instead, one should allow for an acceptable amount of error. The following example checks if `$0 = 5`, with an error of 0.00001 permitted:
+Because computers can only store approximations for most `DOUBLE` and `FLOAT` type values, testing these data types for exact equality or inequality is not recommended.
+Instead, one should allow for an acceptable amount of error.
+The following example checks if `$0 = 5`, with an error of 0.00001 permitted:
```
alexthomasTG: Maybe instead of using =, since that's explicitly not what's happening, say "The following example checks whether `$0` is functionally equivalent to `5` by seeing if `$0` falls in a range of 5 plus or minus 0.0001.

PR #45, opened 2022-05-05
fixed loading job

alexthomasTG:
Updated with correct loading job from previous version and formatted according to style guide


lennessyy (commented):
I'm not sure this is the fix that was requested - I think the wrong file was included here

alexthomasTG:
The two different loading job files were created and referenced correctly, but the one-level flatten loading job was duplicated into both of them. Two different gsql files with identical contents. I took the two-level loading job from an earlier version, verified that it was the one that was being referenced in the docs, and used it in the file that was being referenced.

lennessyy (approved):

PR #46, opened 2022-05-11
add note about privileges

lennessyy:


modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -14,6 +14,11 @@ There are two ways to modify these files to add user-defined functions to GSQL:
 
 * Store the files in a GitHub repository, and configure GSQL to read from the repository.
 * Use `GET` and `PUT` commands to download, modify, and store the files locally.
+** The `GET` command requires the xref:tigergraph-server:user-access:access-control-model.adoc#_privileges[`WRITE_FILE` privilege].
+** The `PUT` command requires the `READ_FILE` privilege.
```
mingxiw: PUT means write. So, should it be "WRITE_FILE" privilege for PUT and "READ_FILE" privilege for GET command?

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -14,6 +14,11 @@ There are two ways to modify these files to add user-defined functions to GSQL:
 
 * Store the files in a GitHub repository, and configure GSQL to read from the repository.
 * Use `GET` and `PUT` commands to download, modify, and store the files locally.
+** The `GET` command requires the xref:tigergraph-server:user-access:access-control-model.adoc#_privileges[`READ_FILE` privilege].
+** The `PUT` command requires the `WRITE_FILE` privilege.
```
mingxiw, line 18: why the READ_FILE has reference, and WRITE_FILE does not have xref?

lennessyy: The reference is to the privileges section since there are no separate individual headings for the two privileges. 
If there were two references, they would point to the same place, so I only added one

mingxiw (approved):

PR #50, opened 2022-05-25
updated guide and pattern matching tutorial

alexthomasTG:


modules/appendix/pages/gsql-style-guide.adoc
```diff
@@ -179,6 +179,17 @@ CREATE QUERY active_members(INT activity_threshold) FOR GRAPH Social_Net
...
+
+Spacing in pattern notation does not affect the output.
+
+`SELECT v FROM start:v -(Posted>:e)- Post:tgt`
+
+is the same as
+
+`SELECT v FROM start:v - (Posted>:e) - Post:tgt`
+
+The first style, with the parentheses connected to the dashes, is preferred.
```
lennessyy, line 192: If we are just specifying the style, then we should just describe the style and give an example. We don't need to say that spacing doesn't affect output - that's describing the syntax.

lennessyy (approved):

modules/appendix/pages/gsql-style-guide.adoc
alexthomasTG: My thinking here was that the user will likely see examples that we haven't caught, either in our docs or in training videos or old examples. That's why I wanted to basically say "you may see this other style, both are fine." Would it be obvious to a new user that the spacing wouldn't affect anything?

PR #53, opened 2022-07-12
remove string compress

lennessyy:
- Remove mention of STRING COMPRESS from the latest documentation


alexthomasTG (approved):
LGTM, no grammar problems introduced by removing. On "Data Types" should there be a note saying "STRING COMPRESS was removed from the documentation for versions 3.6 and later due to its deprecation. Please consult an earlier version of the documentation for information on the STRING COMPRESS data type."

PR #59, opened 2022-08-19
add note for query optimizer

lennessyy:
* Add option for query optimizer


modules/querying/pages/query-operations.adoc
```diff
@@ -577,7 +577,17 @@ Installing takes several seconds for each query. The current version does not su
...
+When installing a query, you have the option of using the query optimizer.
+The query optimizer can potentially reduce the amount of memory a query requires and the time it takes to finish execution.
+The optimizer works by automatically finds the most efficient traversal plan when compiling the query during query installation.
+You can benefit from this option if you query uses a `FROM` clause that performs any number of hops.
+
+To use the query optimizer, use the `-COST` option when installing a query.
+This option can be used together with other options, but it must come after the other options.
+For example, you cannot run `INSTALL QUERY ALL -COST -FORCE`, it must be `INSTALL QUERY ALL -FORCE -COST`
+
+=== Options
```
ainurt, line 581: ```can potentially reduce the amount of memory a query```

We do not optimize to reduce memory consumption (not with this release).

ainurt: Here it is not clear that this is a cost-based query optimizer.

Cost-based optimizer needs a statistics to have accurate estimation in order to find the good plans. (this does not mean cost-optimizer crashes without stats. It will function without but just better to have them to get good quality plans).

I think query optimizer needs it's own page (or sub-page), with more info. To explain that there is also a session variable to enable optimizer. To talk about statistics gathering and how to do it. Similarly to how conjunctive patterns, per clause, and etc advanced features have more of a dedicated pages. 


And here for install option we can keep what you wrote, but also refer user for a cost opt page for full info.

ainurt: Regarding stats: Stats has cardinality (vertex, edge counts) and histograms.  i'd recommend for users to collect refined edge counts (at least) and build histograms on attributes that will be used in WHERE conditions.

ainurt: and by "dedicated page" and "full info" i just meant: all the necessary info to start using cost-optimizer.

lennessyy: Hi Ainur, thanks for the comment. I spoke with Joshua and will add to the doc to reflect the whole process of how to use the optimizer. It was not clear to me before in the QA doc.

modules/querying/pages/query-operations.adoc
```diff
@@ -577,7 +577,17 @@ Installing takes several seconds for each query. The current version does not su
 Concurrent `INSTALL QUERY` commands are allowed as long as only one `INSTALL QUERY` command is running on a single graph. Concurrent `INSTALL QUERY` commands are _not_ allowed on a single graph.
 ====
 
-=== Options for `INSTALL QUERY`
+=== Query optimizer (preview)
+When installing a query, you have the option of using the query optimizer.
+The query optimizer can potentially reduce the amount of memory a query requires and the time it takes to finish execution.
+The optimizer works by automatically finds the most efficient traversal plan when compiling the query during query installation.
```
joshua-tg: I would change this line to something more like this
```
The optimizer uses various statistics to generate an efficient traversal plan when compiling a query.
```
We don't want to say unequivocally that the optimizer will produce a globally optimal plan because it may not. Its main purpose is to produce good plans in general and bad plans never.

modules/querying/pages/query-operations.adoc
```diff
@@ -577,7 +577,17 @@ Installing takes several seconds for each query. The current version does not su
...
+You can benefit from this option if you query uses a `FROM` clause that performs any number of hops.
```
joshua-tg: This line seems a little redundant.

modules/querying/pages/query-operations.adoc
```diff
@@ -596,6 +606,16 @@ a|`-SINGLE`
 |Enables a new internal framework for compiling non-distributed queries. Equivalent to the `single_gpr` session parameter described on the xref:ddl-and-loading:system-and-language-basics.adoc#_session_parameters[Session Parameters page.]
 
 The loop variable in a `xref:querying:control-flow-statements.adoc#_foreach_statement[FOREACH` loop] with this option enabled is treated as a copy, while normally, changes to the loop variable will change the value in the set or bag expression.
+
+a|`-COST`
+|Installs and optimizes the query using the query optimizer.
+This option automatically finds the most efficient traversal plan for executing the query and can reduce the time and memory it takes to execute a query.
+You can benefit from this option if you query uses a `FROM` clause that performs any number of hops.
```
joshua-tg: Again let's not say that the optimizer will produce the most efficient plan. I think it's enough to say that the query optimizer can reduce the time and memory it takes to execute a query.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+Vertex and edge types:
+
+* Any vertex type referenced in the path pattern of a `FROM` clause.
+* Any _refined edge type_.
+A _refined edge type_ means an edge type with a specified pair of `from` and `to` vertex types.
+
+Attributes:
+
+* Any attribute referenced in a xref:select-statement/index.adoc#_where[`WHERE` clause] in a `SELECT` statement in the query.
```
joshua-tg: attribute -> vertex attribute

Edge histogram support won't be available until 3.8

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,96 @@
...
+:description: Steps to enable the cost-based optimizer.
+
+This page walks you through the steps to enable the cost-based optimizer.
+
+== Before you begin
+* You have created a query that contains a xref:select-statement/index.adoc[`SELECT` statement] that uses a xref:select-statement/index.adoc#_path_pattern[path pattern] in its `FROM` clause with one or more hops.
+* You have the privilege to install queries on the graph where you are operating.
+
+== Procedure
+The following is the procedure to enable cost-based optimizer when installing a query.
```
joshua-tg: cost-based optimizer -> cost-based optimization

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+=== Identify referenced types and attributes
+You need to identify the vertex and edge types for which you need to compute cardinality data, and attributes of those types for which you need to compute histogram data.
```
joshua-tg:
```
Identify the vertex and edge types as well as the attributes for which you need to compute statistics.
```
I might also add here that if a user is not sure which statistics to compute, the optimizer will produce a helpful recommendation if it sees any types/attributes don't have statistics computed for them.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,96 @@
...
+We would need the following statistics:
+
+* Cardinality
+** Vertex types
+*** `Tag`
+*** `Comment`
+*** `Person`
+** Refined edge types
+*** `HAS_TAG`: from `Comment` to `Tag`
+*** `LIKES`: from `Person` to `Comment`
```
joshua-tg: @ainurt should we also recommend computing statistics for any reverse edges?

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,96 @@
...
+* Histogram
+** `gender` attribute of type `Person`
+** `id` attribute of type `Tag`
+
+
+=== Compute cardinality data
+To compute cardinality data, use the `POST :14240/gsqlserver/gsql/stats/card` endpoint.
```
joshua-tg: cardinality data -> count statistics (just in case cardinality data is not clear)

Also should we include the default port 14240 here? I image it's pretty easy to extrapolate to a remote instance, but just thought I'd call it out.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+For example, to compute the cardinality data of the vertex `Person` in graph `ldbc_snb`, make the following request:
```
joshua-tg, line 68: cardinality data -> count statistics

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+[source.wrap,console]
+----
+curl -s --user tigergraph:tigergraph -X POST "localhost:14240/gsqlserver/gsql/stats/card?graph=ldbc_snb&vertex=Person"
+----
+
+For another example, to compute the cardinality data of the refined edge type `LIKES` from a `Person` vertex to a `Comment` vertex, make the following request:
```
joshua-tg, line 75: cardinality data -> count statistics

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+[.wrap,console]
+----
+curl -s --user tigergraph:tigergraph -X POST "localhost:14240/gsqlserver/gsql/stats/card?graph=ldbc_snb&edge=LIKES&from=Person&to=Comment"
+----
+
+You need to compute the cardinality data for every type you identified in the previous step, each with its own request.
```
joshua-tg: Instead of saying `You need to`, I would say something like
```
In order to get the best results, it is recommended to compute cardinality statistics for every vertex and edge type referenced in the query.
```
Also I don't know that we need to call out that each computation requires a separate request here. I think it should be clear given the API parameters.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,96 @@
...
+To compute histogram data, use the `POST :14240/gsqlserver/gsql/stats/histogram` endpoint.
+
+For example, to compute the histogram data for attribute `gender` of vertex type `Person` on graph `ldbc_snb`, make the following request:
+
+[.wrap,console]
+----
+$ curl -s --user tigergraph:tigergraph -X POST "http://localhost:14240/gsqlserver/gsql/stats/histogram?graph=ldbc_snb&vertex=Person&attribute=gender&buckets=256"
+----
+
+You need to compute the histogram data for every vertex attribute referenced in a `WHERE` clause of a `SELECT` statement.
```
joshua-tg: Same recommendation here.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,96 @@
...
+=== Install query with `-cost` option
+
+Now that the relevant statistics are available to the query optimizer, you can install the query using the cost-based query optimizer option.
+To use the cost-based optimizer during installation, use the `-cost` option when installing the query.
+
+For example, to install the query `example3`, run `INSTALL QUERY example3`.
+
+Alternatively, set the session variable to true by running `set cost_opt = true` in the GSQL shell.
+
+If you missed any cardinality or histogram data, GSQL warns you about the missing data, so you can compute the data for any missing type or attribute.
```
joshua-tg: Ah I see you called this out here. I might call this out at the beginning just so that the user knows right away that they can let the optimizer make this determination.

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,34 @@
+= Query Optimizer
+:description: Overview of the query optimizer.
+
+GSQL has a built-in query optimizer that can optimize the execution plan of a query when you install the query.
+Currently, the optimizer can optimize the execution plan in the following ways:
+
+* <<_cost_based_optimization>>
+
+== What is the query optimizer?
```
joshua-tg: I think we can merge this sub-header into the main `Query Optimizer` header.

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,34 @@
...
+The query optimizer optimizes the compiled binary execution plan of a query.
+
+When you xref:query-operations.adoc#_install_query[install a GSQL query], GSQL compiles the query logic into a binary execution plan.
+If your query contains a path pattern, the pattern contains a source vertex set, one mor more path edge patterns, and one or more step vertex sets.
+Similar to the path pattern in GSQL, its binary execution plan also starts from a set of source vertices, traverses through edges in the specified pattern, and finds the target vertices.
+
+However, the actual execution plan of the query might perform path traversals differently than specified in the GSQL query to product results.
+When you install a query, the query optimizer can optimize the execution plan of a query so that it executes more efficiently.
+
+Depending on the type of optimization used, this could allow a query to execute with less time, less memory, or less storage.
```
joshua-tg: There's a lot of good info in this section, but it might be too much. I think I would try to condense it. For instance, I think it would be sufficient to say something like
```
There may be many equivalent ways to express the same query in GSQL.
The job of the optimizer is to find the best way, or more accurately, avoid the most inefficient execution plans.
As a result, the user queries for the data they're interested in but does not have to worry about the details of how it's retrieved.
```

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+[#_cost_based_optimization]
+== Cost-based optimization
+
+Cost-based optimization reduces the number of edge traversals a query has to execute.
+Enabling the cost-based optimizer when installing a query has the potential to significantly reduce the execution time of the installed query.
+
+=== How it works
+The cost-based optimizer takes statistics regarding the vertex and edge types as well as their attributes that are referenced in the query to compile an execution plan that involves the fewest edge traversals.
```
joshua-tg: It's true that right now we're mostly trying to minimize the number of edge traversals, but with time our cost function will have more parameters. Do we want to call this out explicitly in the documentation? What do you think @ainurt?

modules/querying/pages/query-optimizer/stats-api.adoc
```diff
@@ -0,0 +1,19 @@
...
+== Cardinality
+
+The following endpoints compute, fetch and delete vertex and edge counts (cardinality) of a graph.
+
+
+include::partial$cardinality-api.adoc[]
+
+== Histogram
+
+Histogram is a type of attribute statistics that stores information about the distribution of values of an attribute.
```
joshua-tg:
```
Histograms store information about the distribution of attribute values.
```

modules/querying/partials/cardinality-api.adoc
```diff
@@ -0,0 +1,109 @@
...
+|Parameter |Description |Type
+
+|`graph`
+|Name of the graph.
+Required.
+|`STRING`
+
+|`vertex`
+|Vertex type to compute statistics for.
+You must provide either the `vertex` or `edge` parameter.
```
joshua-tg, line 22: But not both.

modules/querying/partials/cardinality-api.adoc
```diff
@@ -0,0 +1,109 @@
...
+|`STRING`
+
+|`edge`
+|Vertex type to compute statistics for.
```
joshua-tg: Vertex -> Edge

modules/querying/partials/cardinality-api.adoc
```diff
@@ -0,0 +1,109 @@
...
+You must provide either the `vertex` or `edge` parameter.
+If you provided it an edge type, you also need to provide the `from` and `to` parameters.
+|`STRING`
+
+|`from`
+|Vertex type of the  `from` endpoint of an edge type.
```
joshua-tg: Maybe
```
The source vertex type of an edge.
```
instead.

modules/querying/partials/cardinality-api.adoc
```diff
@@ -0,0 +1,109 @@
...
+Required if `edge` is provided.
+|`STRING`
+
+|`to`
+|Vertex type of the  `from` endpoint of an edge type.
```
joshua-tg: Maybe
```
The target vertex type of an edge.
```

modules/querying/partials/histogram-api.adoc
```diff
@@ -0,0 +1,127 @@
...
+Required.
+|`STRING`
+
+|`bucket`
+|The number of intervals (https://en.wikipedia.org/wiki/Data_binning[buckets]) to divide the range of attribute values by.
+Required.
+
+It is recommended that each bucket has at least 100 values.
+Therefore, a good rule of thumb is to divide the number of vertices by 100 to decide the number of buckets.
+For example, if the vertex type has 100,000 vertices, set this parameter to 1000.
```
joshua-tg: I would say at least 100 values. So instead of recommending that you set the parameter to `1000` exactly, we should recommend to set it no greater than `1000`. And then we should probably add that 256 is typically a standard default.

modules/querying/partials/histogram-api.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+|`graph`
+|Name of the graph.
+Required.
+|`STRING`
+
+
+|`vertex`
+|Name of the vertex type.
+Required.
```
joshua-tg: Technically vertex is optional. Only required if an attribute is specified.

modules/querying/partials/histogram-api.adoc
```diff
@@ -0,0 +1,127 @@
...
+|`STRING`
+
+|`attribute`
+|Name of the attribute.
+Required.
```
joshua-tg: Attribute is optional.

modules/querying/partials/histogram-api.adoc
```diff
@@ -0,0 +1,127 @@
...
+|`STRING`
+|===
+
+==== Examples
+
+The following request deletes the histogram statistics for the graph `ldbc_snb`:
+
+[.wrap,console]
+----
+curl -s --user tigergraph:tigergraph -X DELETE "http://localhost:14240/gsqlserver/gsql/stats/histogram?graph=ldbc_snb=
```
joshua-tg: You added a stray `=` at the end. We should also add an example for the vertex and attribute variants.

joshua-tg (changes requested):

modules/querying/pages/query-operations.adoc
ainurt: I'd remove part about reducing memory usage, we do not currently optimize for memory. only for performance.

ainurt: also, can we just say path patterns. i think we do use term "path patterns" in our user docs. just need to make sure it does not include conjunctive patterns.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
ainurt: do you mean like "HAS_TAG_REVERSE",  suppose that's the name reverse edge is defined with?

since refined edge count will not be computed automatically for the reverse edge as well when computing HAS_TAG, we should.

modules/querying/pages/query-optimizer/index.adoc
ainurt: at least for now we can probably omit the details of our cost-model ?!

modules/querying/pages/query-optimizer/index.adoc
ainurt: here again, i would not say that we explicitly optimize for less memory or less storage.

modules/querying/pages/query-operations.adoc
lennessyy: I do say that depending on the type of optimization used it could improve performance in different ways, so they can expect different kinds of optimizations in the future;

I'll change the verbiage to path patterns

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
lennessyy: We should include the default because many of the RESTPP endpoints are on port 9000 and people are used to most REST endpoints being on 9000. We have too many ports for our REST API.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
lennessyy: I will add "for best results" to the end of the sentence. We want to be concise when we can; 

As for emphasizing that each computation requires its own request, I felt like it's good to have there because I was not sure about that from reading the parameters and had to ask you. 

If I was a user and had a very long query with dozens of types, which I likely have if I am using the optimizer, I would probably wonder if I need to make that many requests - it can sound a little bad, but it's worse if they have to wonder and then find out.

modules/querying/pages/query-optimizer/index.adoc
lennessyy: Since there will be more types added in the future, i want the different options to be at the top so it's easy to navigate. It also helps to open with a relatively simple description, and the delve a bit deeper for a detailed description later when we have something as complex as this.

modules/querying/pages/query-optimizer/index.adoc
lennessyy: The reason I felt the need to call this out is that as a user, this was the only tangible way i was able to understand what "cost" means. Cost could mean many different things to people - people could understand it as memory or other resources since it is a broad term in its daily usage. 

However, if we feel strongly that this is a confidential detail that we don't want to disclose, I can just say "compile an execution plan that takes the least time."

modules/querying/pages/query-optimizer/index.adoc
lennessyy: Josh: I usually agree with supplying less information on things like this, but my sense is that most of our users do not know what `INSTALL QUERY` actually does, and do not know that if you write a query differently, it might be more or less efficient to begin with. For them to even begin to understand why writing a query differently would yield different performance, I feel we need to go a bit into the details. 

Ainur: I will add a sentence after to say that we currently only support cost-based optimizer. This paragraph is about the optimizer as a whole and what it has the potential to do, so it isn't limited to the cost-based optimizer.

modules/querying/partials/cardinality-api.adoc
```diff
@@ -0,0 +1,110 @@
+[#_compute_cardinality_statistics]
+=== Compute cardinality statistics
+
+`POST :14240/gsqlserver/gsql/stats/card`
+
+This endpoint computes the cardinality statistics for the vertex and edge types specified in the request.
+The cardinality statistics describes the number of vertices of a specified type, and the number of edges of a specified type between a specified pair of vertex types.
```
alexthomasTG: describe

modules/querying/partials/cardinality-api.adoc
```diff
@@ -0,0 +1,110 @@
...
+[.wrap,console]
+----
+curl -s --user tigergraph:tigergraph -X POST "localhost:14240/gsqlserver/gsql/stats/card?graph=ldbc_snb&edge=LIKES&from=Person&to=Post"
+----
+
+=== Retrieve cardinality statistics
+
+`GET :14240/gsqlserver/gsql/stats/card`
+
+This endpoint retrieves cardinality statistics of a graph.
```
alexthomasTG: retrieves the cardinality statistics

alexthomasTG (approved):

modules/querying/pages/query-operations.adoc
```diff
@@ -596,6 +596,16 @@ a|`-SINGLE`
 |Enables a new internal framework for compiling non-distributed queries. Equivalent to the `single_gpr` session parameter described on the xref:ddl-and-loading:system-and-language-basics.adoc#_session_parameters[Session Parameters page.]
 
 The loop variable in a `xref:querying:control-flow-statements.adoc#_foreach_statement[FOREACH` loop] with this option enabled is treated as a copy, while normally, changes to the loop variable will change the value in the set or bag expression.
+
+a|`-COST`
+|Installs and optimizes the query using the xref:query-optimizer/index.adoc#_cost_based_optimization[cost-based query optimizer].
+The optimizer uses various statistics to generate an efficient traversal plan when compiling a query.
+You can benefit from this option if you query uses a `FROM` clause that uses a xref:select-statement/index.adoc#_path_pattern[path pattern].
```
joshua-tg: you -> your

modules/querying/partials/histogram-api.adoc
```diff
@@ -0,0 +1,130 @@
...
+|`attribute`
+|Name of the attribute.
+Required.
+|`STRING`
+
+|`bucket`
+|The number of intervals (https://en.wikipedia.org/wiki/Data_binning[buckets]) to divide the range of attribute values by.
+Required.
+
+It is recommended that each bucket has at least 100 values.
```
joshua-tg: has -> have

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,34 @@
+= Query Optimizer
+:description: Overview of the query optimizer.
+
+GSQL has a built-in query optimizer that can optimize the execution plan of a query when you install the query.
+As of TigerGraph 3.7, the optimizer can optimize a query in the following ways:
+
+* <<_cost_based_optimization>>
```
joshua-tg: I would remove this bullet just because the optimizer is really synonymous with cost-based optimization. So I doubt we'd ever add another point to this list.

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+== What is the query optimizer?
+The query optimizer optimizes the compiled binary execution plan of a query.
+
+When you xref:query-operations.adoc#_install_query[install a GSQL query], GSQL compiles the query logic into a binary execution plan.
+If your query contains a xref:querying:select-statement/index.adoc#_path_pattern[path pattern], the pattern contains a source vertex set, one or more path edge patterns, and one or more step vertex sets.
```
joshua-tg: nit: not a sentence.

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,34 @@
...
+Similar to the path pattern in the GSQL query, its binary execution plan also starts from a set of source vertices, traverses through edges, and finds the target vertices.
+
+However, the actual execution plan of the query might perform path traversals differently than specified in the GSQL query to produce results.
+When you install a query, the query optimizer can optimize the execution plan of a query so that it executes more efficiently.
+
+Depending on the type of optimization used, this could allow a query to execute with less time, less memory, or less storage.
+As of 3.7, TigerGraph supports cost-based optimization, which reduces the execution time of a query.
```
joshua-tg: I would reword this sentence. To me this implies that cost-based optimization will always reduce execution time, and that some other technique will be need to optimize for memory, etc. But cost-based optimization is a general optimization framework that can be used to optimize across any dimension including execution time, memory, or resource utilization in general. As of this release, we are optimizing solely for execution time, but we will be optimizing across other dimensions in future releases.

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+[#_cost_based_optimization]
+== Cost-based optimization
+
+Cost-based optimization reduces time it takes to execute a query.
+
+
+=== How it works
```
joshua-tg: `How it works` and `What is the query optimizer?` seem to be saying the same thing. Do we need both?

joshua-tg (changes requested):

modules/querying/pages/query-optimizer/index.adoc
lennessyy: How it works is under the cost-based optimizer, so it only describes how the cost-based optimizer works. I assume that the other types of optimizations are going to work differently?

modules/querying/pages/query-optimizer/index.adoc
lennessyy: hmm, which one isn't a sentence?

modules/querying/pages/query-optimizer/index.adoc
lennessyy: ahh I see. from Ainur's description I thought there'd be other kinds of optimization such as "memory-based" or something, but we won't have those?

joshua-tg: We don't optimize for memory yet, but when we do, it will still be considered cost-based optimization.

modules/querying/pages/query-optimizer/index.adoc
joshua-tg: Oh I think I misread it. Are you defining what a path pattern is? I thought the sentence was missing a predicate. I find it a little confusing since this is the only reference to path patterns.

modules/querying/pages/query-optimizer/index.adoc
alin-tigergraph: There is a strategic issue with this intro. 
We do not want to trumpet to the world that we did not have an optimizer before 3.7.
As a matter of fact we had one but it was heuristics-based.

So I would rephrase as follows:

As of TigerGraph 3.7, the GSQL Syntax version V2 is the default. This syntax supports multi-hop path patterns in the FROM clause. Path patterns match a potentially unbounded sequence of edges. The path patterns are declarative, specifying the structure of paths to be matched, without specifying the particulars of how the graph is traversed to find these paths. TigerGraph's cost-based optimizer chooses traversal plans automatically and transparently to the user.

modules/querying/pages/query-optimizer/index.adoc
joshua-tg: Not quite. We'll add further improvements and new dimensions to the cost model, but they'll all be driven by the same underlying algorithm, if that makes sense.

alin-tigergraph:
Thanks, Lenny, for this work!

All, I am a bit of a clutz regarding the github review interface.
I have added a comment, is it visible?
Thanks,

--alin

modules/querying/pages/query-optimizer/index.adoc
joshua-tg: Thanks @alin-tigergraph that's a good point.

modules/ddl-and-loading/pages/system-and-language-basics.adoc
```diff
@@ -267,6 +267,8 @@ This session parameter does not affect syntactic errors: GSQL will always exit o
 a| `single_gpr` | Changes the internal engine framework for single queries to the engine framework currently used for distributed queries, also known as GPR. Enabling this mode may increase performance in certain cases. Default value is `false`.
 
 The loop variable in a `xref:querying:control-flow-statements.adoc#_foreach_statement[FOREACH` loop] with this option enabled is treated as a copy, while normally, changes to the loop variable will change the value in the set or bag expression.
+
+|`cost-opt`| Whether to use the xref:querying:query-optimizer/index.adoc[query optimizer] when  installing a query. If set to true, query installation during this session uses the optimizer to optimize the query based on cost.
```
joshua-tg: Since we're calling it the query optimizer vs the cost-based optimizer, can we remove the second sentence here?

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,108 @@
+= Enabling Query Optimizer
```
joshua-tg: To @alin-tigergraph's point, maybe we want to call this section `Aiding the Query Optimizer`. A sophisticated user will be able to determine that this is a new feature, but we probably don't want to be overt about it.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,108 @@
...
+:sectnums:
+:description: Steps to enable the query optimizer.
```
joshua-tg, line 3: Similarly here, maybe we should update the description like so
```
How to help the query optimizer in choosing optimal plans.
```

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,108 @@
...
+$ curl -s --user tigergraph:tigergraph -X POST "http://localhost:14240/gsqlserver/gsql/stats/histogram?graph=ldbc_snb&vertex=Person&attribute=gender&buckets=256"
+----
+
+For best results, compute the histogram data for every vertex attribute referenced in a `WHERE` clause of a `SELECT` statement.
+
+[#_install_query_with_cost_option]
+=== Install query with `-cost` option
+
+Now that the relevant statistics are available to the query optimizer, you can install the query using the cost-based query optimizer.
+To use the cost-based optimizer during installation, use the xref:query-operations.adoc#_install_query[`-cost` option] when installing the query.
```
joshua-tg: We probably want to remove references to cost-based here.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+For example, to install the query `example3`, run `INSTALL QUERY example3 -cost`.
+
+Alternatively, set xref:ddl-and-loading:system-and-language-basics.adoc#_session_parameters[the `cost_opt` session parameter] to true by running `set cost_opt = true` in the GSQL shell.
+This enables the query optimizer for the rest of your GSQL session.
+You no long need to supply the `-cost` option when installing queries to use the optimizer if the `cost_opt` parameter is set to true.
+
+If you missed any cardinality or histogram data, the optimizer warns you about which types are missing, so you can compute the data for any missing type or attribute.
```
joshua-tg, line 108: Note: this looks like a repeat of line 33, so you can probably remove this sentence.

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,24 @@
+= Query Optimizer
+:description: Overview of the query optimizer.
+
+NOTE: This is a Preview feature and should not be used for production purposes.
+
+The query optimizer optimizes the execution plan of a query based on cost.
+
+When you write a GSQL query with a xref:querying:select-statement/index.adoc#_path_pattern[path pattern (Syntax V2)], you specify a pattern that contains a source vertex set, one or more path edge patterns, and one or more step vertex sets.
+This pattern is declarative.
```
joshua-tg: Can these two sentences be condensed?
```
When writing a multi-hop GSQL query in v2 syntax, the pattern specified in the FROM clause is declarative.
```

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,24 @@
...
+In other words, it specifies the structure of paths to be matched, without specifying the particulars of how the graph is traversed to find these paths.
```
joshua-tg, line 7: Can you add a sentence directly after this one saying
```
It is the query optimizer's job to determine the best graph traversal algorithm to satisfy the query.
```

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+When you xref:query-operations.adoc#_install_query[install the GSQL query], GSQL compiles the query logic into a binary execution plan.
+The query optimizer takes statistics such as the number of instances for each vertex and edge type, as well as the distribution of values in the attributes relevant to the query.
+It then optimizes the execution plan to traverse the graph based on those statistics to minimize the cost of executing the query.
```
joshua-tg: I think we should merge this paragraph with the previous one. I don't think we need the first sentence. I think it's prior knowledge that we can assume the user already has if they're reading this, since it's called out in other areas of the docs.

I also don't think we need to be explicit about the various statistics here. You've been explicit about those elsewhere, so I think it's fine to say
```
It is the query optimizer's job to determine the best graph traversal algorithm to satisfy the query.
It does so by utilizing various pre-computed statistics describing the data distribution of the types referenced in the query.
It then optimizes the execution plan to traverse the graph based on those statistics to minimize the cost of executing the query.
```

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+As of 3.7, the Preview version of the optimizer is focused on reducing query execution time.
```
joshua-tg: You already mentioned that this is a preview feature, so I think this should just be
```
Currently the optimizer is focused on reducing overall query execution time.
```

modules/querying/pages/query-optimizer/index.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+NOTE: The query optimizer has a more significant effect when more hops that a query has to perform are combined in a single multi-hop `SELECT` statement using path patterns in Syntax V2, as opposed to using multiple one-hop `SELECT` statements.
```
joshua-tg: I think this can be removed if you update line 8 to included multi-hop and v2 syntax. If we make sure that the context is multi-hop v2 syntax, then we don't even need to mention v1.

modules/querying/partials/cardinality-api.adoc
```diff
@@ -0,0 +1,110 @@
+[#_compute_cardinality_statistics]
+=== Compute cardinality statistics
+
+`POST :14240/gsqlserver/gsql/stats/card`
+
+This endpoint computes the cardinality statistics for the vertex and edge types specified in the request.
+The cardinality statistics describe the number of vertices of a specified type, and the number of edges of a specified type between a specified pair of vertex types.
```
joshua-tg, line 7: One thing I forgot to mention - can you call out that this API is a preview feature and subject to change in the future? Just so that it's clear that these APIs should not be used in production.

modules/querying/partials/histogram-api.adoc
```diff
@@ -0,0 +1,130 @@
+[#_compute_histogram_statistics]
+=== Compute histogram statistics
+
+`POST :14240/gsqlserver/gsql/stats/histogram`
+
+This endpoint computes histograms for a specified attribute of a vertex type.
```
joshua-tg, line 6: Similarly here, can you call out that this is a preview feature and subject to change.

joshua-tg (changes requested):
I added some more comments. Mainly just to condense some of the documentation, and to be a little less overt about advertising the query optimizer as a new feature.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
lennessyy: It's quite normal to repeat information wherever it's relevant. Users never read the docs from beginning to end - we just want them to see the information wherever it could possibly be relevant.

modules/querying/pages/query-optimizer/index.adoc
lennessyy: We cannot assume everyone reading this page has this knowledge. The best practice for information architecture for online documentation which we (try to) follow is called Every Page is Page One. People can stumble on the page from many sources for any reason - we should always try to provide references to background information they need to understand the page.

modules/querying/pages/query-optimizer/index.adoc
lennessyy: Again, I don't think it hurts to repeat this information here.

modules/querying/partials/cardinality-api.adoc
lennessyy: I added to the stats API page that all the APIs on the page are Preview

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
lennessyy: I don't think we are being overt about this feature. It is a new feature, and we are introducing it like any other new feature. "Aiding Query Optimizer" complicates the subject for users because before 3.7, the term Query Optimizer was never used, which makes sense, because we are using it strictly in the sense of cost-based optimization.

modules/ddl-and-loading/pages/system-and-language-basics.adoc
lennessyy: "optimize the query based on cost" is describing what the query optimizer does. I'll reword it a little bit to make sure that any implication that there might other types of optimization is not present, but it is helpful to have a brief description of what the option does.

modules/querying/pages/query-optimizer/enable-cost-optimizer.adoc
lennessyy: removed.

modules/querying/pages/query-optimizer/index.adoc
lennessyy: We do mention conceptually how the optimizer works, but users still might not be able to connect the dots if we do not tell them explicitly how they need to write their queries.

alin-tigergraph (approved):

alin-tigergraph:
Very nice job, Lenny, thank you!

I agree with Josh's change requests, but I approve so this doesn't block on me.

joshua-tg (approved):
lgtm!

PR #63, opened 2022-09-09
add clarification for like backslash

lennessyy:
- Add example for`LIKE` operator


alexthomasTG (approved):
Excellent explanation

quanchentg (approved):

PR #65, opened 2022-09-24
add clarification for accum

lennessyy:


modules/appendix/examples/ldbc_snb/select_statement_accum.gsql
```diff
@@ -0,0 +1,11 @@
+CREATE QUERY accum_wrong_example() SYNTAX v2 {
+    SumAccum<INT> @@count_total;
+    SumAccum<INT> @active_flag = 0;
```
pengwangTigerGraph, line 3: @lennessyy Hi, originally I set ```@active_flag``` to be ```MaxAccum<INT>```, as ```MaxAccum<INT> @activeFlag = 0;```. Because I want to cap the value of ```@active_flag``` to be 1, each time the where clause is satisfied. Then ```@@countTotal``` will be invoked together with ```@active_flag```, so every time where clause is satisfied, ```@@countTotal``` will also increase by 1, since ```@activeFlag``` is MaxAccum with maximum as 1. This is not a good practice, I only made it up to show the unexpected result from  reusing updated variables (on the right hand side) within the same ACCUM clause.

alexthomasTG (approved):

modules/appendix/examples/ldbc_snb/select_statement_accum.gsql
lennessyy: thanks for the explantion. I don't think we need to stress about this as this isn't what the example is trying to show; we are just trying to show that updating variables in an ACCUM clause is not a good idea., which is accomplished by the example

PR #77, opened 2022-10-27
Doc 1501 udf update

lennessyy:


modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -75,22 +66,42 @@ inline string reverse(string str){
...
+There are three ways to modify these files to add user-defined functions to GSQL:
+
+* Store UDF files in a GitHub repository, and configure GSQL to read from the repository.
+** This is the recommended approach.
+It also takes the highest precedence.
+If GSQL is configured to read from GitHub for UDFs, and the configurations are valid, local UDF files are ignored.
+* Use `gadmin` commands to upload a local UDF file to GSQL.
+** Use this approach when your server does not have access to the internet and cannot access GitHub.
+* Use `GET` and `PUT` commands to download, and upload the files directly through GSQL shell.
+** For security reasons, this method is disabled by default.
```
romansod: @mingxiw @qianzhang-tg For the official documentation let's make sure we get the exact wording that we want here. How we word this is important.

I could see us phrase the new gadmin config, and github, as secure udf file offerings, omitting PUT as a less favorable option. 

We can add a footnote or * or something to select github and gadmin config and indicate these offerings provide the most secure and convenient way to manage UDFs. Then have a disclaimer on the PUT option that this is disabled by default and should only be used sparingly, with deliberation.

What are your thoughts?

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -3,26 +3,17 @@
...
-This page introduces the process to define a query UDF. Once defined, the new functions are added into GSQL automatically the next time GSQL is executed.
+This page introduces the process to define a query UDF.
+Once defined, the new functions are added into GSQL automatically the next time GSQL is executed.
+
+== UDF files
 
 UDFs are written in C{pp} in two files, `ExprFunctions.hpp` and `ExprUtil.hpp`:
 
 * ExprFunctions is used for functions that are called directly in GSQL queries.
 * ExprUtil contains structs or helper functions that used called by the functions in ExprFunctions.
```
qianzhang-tg: `used called` seems to be a typo.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -126,10 +138,78 @@ gadmin config apply
...
+$ gadmin config set GSQL.UDF.ExprFunctions   @<path_to_file>
+----
+
+
+For example:
+
+[.wrap,console]
+----
+$ gadmin config set GSQL.UDF.ExprFunctions   @/home/tmpExprFunctions.hpp
+----
```
qianzhang-tg: I suggest to also append `gadmin config apply` command in the example similar to the Github example above.
Also, similar to gadmin setting in Github example, specify that users need to restart GSQL if it is still running to make changes work.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -171,87 +251,71 @@ PUT ExprUtil successfully.
...
   x = 3.5;
   PRINT greater_than_three(x);
   y = greater_than_three(2.5);
   PRINT y;
-
 }
 ----
 
+After you are done uploading UDFs, it is strongly recommended that you disable uploading UDFs through GSQL until you need to modify the UDFs again:
+
```
qianzhang-tg, line 213: This part is not correct.
Once the `GSQL.UDF.EnablePutExpr` is set to false and takes effect, GSQL will load contents from `GSQL.UDF.ExprUtil` and `GSQL.UDF.ExprFunctions`. They are not necessarily the same as what user updated with PUT command.

If we want users to turn off `GSQL.UDF.EnablePutExpr` after users are satisfied with experimenting UDF file via PUT command. Then we need to ask them to manually set `GSQL.UDF.ExprUtil` and `GSQL.UDF.ExprFunctions` with the UDF files.

Without this step and just turning off PUT command will make GSQL to load user defined files from previous set `GSQL.UDF.ExprUtil` and `GSQL.UDF.ExprFunctions` values.

qianzhang-tg: Using PUT command update UDF files will not automatically change `gadmin config`'s `GSQL.UDF.ExprUtil` and `GSQL.UDF.ExprFunctions`

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -171,87 +251,71 @@ PUT ExprUtil successfully.
...
+[source.wrap,console]
+----
+$ gadmin config set GSQL.UDF.EnablePutExpr false
+$ gadmin config apply
+$ gadmin restart gsql
+----
+
+
```
qianzhang-tg, line 228: Above example should change accordingly

modules/querying/pages/func/query-user-defined-functions.adoc
lennessyy: I see. I only put this in because I saw this in the design files. I'll remove this paragraph.

modules/querying/pages/func/query-user-defined-functions.adoc
qianzhang-tg: I agree that the documentation should add some visual cues to show which one is desired and which one is not.

modules/querying/pages/func/query-user-defined-functions.adoc
qianzhang-tg: Another notice about switching from PUT command to gadmin is that user should always set `GSQL.UDF.ExprUtil` and `GSQL.UDF.ExprFunctions` before or at the same time when the user sets `GSQL.UDF.EnablePutExpr` to false, apply config and restart gsql.

The case to AVOID is to set `GSQL.UDF.EnablePutExpr` to false, apply config and restart gsql, then set `GSQL.UDF.ExprUtil` and `GSQL.UDF.ExprFunctions`, apply config and restart gsql. The first restart of gsql will use previous set `GSQL.UDF.ExprUtil` and `GSQL.UDF.ExprFunctions` values which may not be what user wants to use. And the unexpected update in UDF files can cause previous installed queries to be uninstalled.

qianzhang-tg: > I see. I only put this in because I saw this in the design files. I'll remove this paragraph.

I am okay with removing the paragraph here. But I do think it is worth to be mentioned somewhere in documentation what user needs to pay attention when they decide to switch between enable/disable PUT command.

modules/querying/pages/func/query-user-defined-functions.adoc
lennessyy: I can try to take more of a pros and cons approach to this - saying that using PUT gives more flexibility as you don't need to have access to the file systems of the server, but it might also introduce security risks, unless we want to be very explicit about its flaws.

arunramasami: Hi Lenny - I recommend focusing on GitHub and gadmin config offering rather than PUT. I don't want us to emphasize that PUT creates security vulnerabilities. We don't want to accept Neo4j's claims. Instead, we will focus on the GitHub and gadmin config as the recommended secured offering & how it is more secure for users/customers. That should be our approach. I can review it once it is finalized. Thanks Lenny and team.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -171,87 +256,63 @@ PUT ExprUtil successfully.
...
 
   x = 3.5;
   PRINT greater_than_three(x);
   y = greater_than_three(2.5);
   PRINT y;
-
 }
 ----
 
+WARNING: If you set `GSQL.UDF.EnablePutExpr` to `false` after uploading UDFs using the `PUT` command, GSQL ignores the functions uploaded through the `PUT` command and reads those functions from `GSQL.UDF.ExprUti` and `GSQL.UDF.ExprFunctions` instead.
```
qianzhang-tg: typo: should be `GSQL.UDF.ExprUtil`

qianzhang-tg (approved):
Looks good to me.

alexthomasTG (approved):

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -75,22 +67,39 @@ inline string reverse(string str){
 }
 ----
 
-== Use GitHub to store UDFs
+== Upload UDFs
+
+There are two ways to modify these files to add user-defined functions to GSQL:
```
romansod, line 73: I think we can say here that these two options are securely managed by gadmin

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -75,22 +67,39 @@ inline string reverse(string str){
...
+
+* Store UDF files in a GitHub repository, and configure GSQL to read from the repository.
+** This is the recommended approach.
+It also takes the highest precedence.
+If GSQL is configured to read from GitHub for UDFs, and the configurations are valid, local UDF files are ignored.
+* Use `gadmin` commands to upload a local UDF file to GSQL.
+** Use this approach when your server does not have access to the internet and cannot access GitHub.
+
+NOTE: There is a legacy method of xref:func/upload-udf-legacy.adoc[uploading] UDFs through GSQL.
+For security reasons, the legacy method is not recommended.
```
romansod: Instead of saying it is not recommended because of security reasons, lets say:

_This method is not managed by gadmin and is therefor not recommended_

modules/querying/pages/func/upload-udf-legacy.adoc
```diff
@@ -0,0 +1,83 @@
+= Uploading User-Defined Functions (Legacy)
+:sectnums:
+
+
+TigerGraph offers the ability to upload UDFs directly from GSQL.
+This allows authorized users to upload UDF files without having to access the server file system.
+
```
romansod: If your host machine is not completely secure, it is strongly recommended that you upload UDFs through `gadmin` instead of the steps on this page.
For this reason, this option is disabled by default and must be turned on by the TigerGraph Linux user.
Once enabled, UDFs are no longer managed by `gadmin` until this option is once again disabled.

modules/querying/pages/func/upload-udf-legacy.adoc
```diff
@@ -0,0 +1,83 @@
...
+For security reasons, this option is disabled by default and must be turned on by the TigerGraph Linux user.
+Once enabled, UDFs are no longer managed by `gadmin`.
+If your host machine is not completely secure, it is strongly recommended that you upload UDFs through `gadmin` instead of the steps on this page.
+
```
romansod: @lennyjiaxichen This suggestion I leave up to you as this is something you know better than me, but I think it might be good to actually include a link _back_ to the recommended options here

qianzhang-tg (approved):
Looks good to me.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -126,12 +131,27 @@ gadmin config apply
 
 After the parameters are successfully configured, you can access your UDFs in new queries right away.
 
+=== Upload UDFs with GSQL `PUT` command
+TigerGraph offers the ability to upload UDFs directly from GSQL.
+This allows authorized users to upload UDF files without having to access the server file system.
+
+For security reasons, this option is disabled by default and must be turned on by the TigerGraph Linux user.
```
romansod: Replace line 138 with something like this:
"This command is securely managed by gadmin and is disabled by default, requiring a Tigergraph Linux user with gadmin access to enable it" 

Replace line 139 with something like this:
"After making modifications to your UDF files, it is recommended that you disable the ability to upload UDF files using PUT until you need to modify the files again"

romansod (approved):
Beautiful, looks good Lenny thanks!

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -75,22 +68,33 @@ inline string reverse(string str){
...
+== Upload UDFs
 
-You can configure GSQL to read from a GitHub repository for ExprFunctions and ExprUtil.
+There are two ways to modify these files to add user-defined functions to GSQL:
 
-If GitHub access is configured, GSQL will retrieve user source code files from GitHub before files added via `PUT`, so long as the files exist.
+* Store UDF files in a GitHub repository, and configure GSQL to read from the repository.
+** This is the recommended approach.
+It also takes the highest precedence.
+If GSQL is configured to read from GitHub for UDFs, and the configurations are valid, local UDF files are ignored.
```
victorleeTG, line 78: 1. Reword: If GSQL is configured to read UDFs from GitHub,...
2. What is a "local UDF file"?  The previous version defined this above when introducing GET and PUT, but you have now deleted that section.

modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -75,22 +68,33 @@ inline string reverse(string str){
...
 
-== Use GitHub to store UDFs
+== Upload UDFs
 
-You can configure GSQL to read from a GitHub repository for ExprFunctions and ExprUtil.
+There are two ways to modify these files to add user-defined functions to GSQL:
 
-If GitHub access is configured, GSQL will retrieve user source code files from GitHub before files added via `PUT`, so long as the files exist.
+* Store UDF files in a GitHub repository, and configure GSQL to read from the repository.
+** This is the recommended approach.
```
victorleeTG, line 76: Where is the security angle to GitHub access?

lennessyy: What do you mean by security angle?

modules/querying/pages/func/query-user-defined-functions.adoc
lennessyy: I'll change this to "UDFs in other sources are ignored." The emphasis here is that if you configured GitHub, GSQL only reads from GitHub.

PR #81, opened 2022-11-18
update with notes on precision limits

alexthomasTG:


qingGongTG (approved):

modules/querying/pages/operators-and-expressions.adoc
```diff
@@ -171,6 +171,15 @@ We support the standard Boolean operators and standard order of precedence: `AND
...
 Bit operators (`<<` `>>` `&` `|`) operate on integers and return an integer.
 
+TigerGraph implements bit operators in the same way that C++ does.
+This returns an undefined result in the following situations:
+
+* If you left-shift a signed number so that the sign bit is affected
+* If the additive expression is negative
+* If the additive expression is greater than or equal to the number of bits in the (promoted) shift expression
+
+See Microsoft's documentation on link:https://learn.microsoft.com/en-us/cpp/cpp/left-shift-and-right-shift-operators-input-and-output[C++ bit shift operators] for more information.
```
junhyungshin-tg: the reference from microsoft might differ since it uses different compiler (msvc vs gcc). I think we need to give another reference, or at least remove this.

junhyungshin-tg (approved):
Thanks Alex. As long as we mention it behaves the same as C++, I think it should be straightforward enough.

PR #82, opened 2022-11-21
add notes about types

alexthomasTG:


modules/querying/pages/func/edge-methods.adoc
```diff
@@ -28,8 +28,14 @@ The data type of the attribute itself.
 | Parameter | Description | Data type
 
 | `attrName`
-| The name of the attribute.
-| `STRING`
+| The name of an attribute. This can be a query parameter, constant string, or global string variable.
+| Base types: `STRING` `INT` `UINT` `DATETIME` `DOUBLE` `FLOAT` `BOOL`
+
+Container types: `LIST<INT/DATETIME/DOUBLE/STRING>`, `SET<INT/DATETIME/DOUBLE/STRING>`
```
yuchenZhangTG: Could you use `CONTAINER<TYPE>` where `CONTAINER` can be `LIST` or `SET` and `TYPE` can be `INT`, `DATETIME`, `DOUBLE`, `STRING`. For example `LIST<INT>`

modules/querying/pages/func/vertex-methods.adoc
```diff
@@ -182,8 +182,13 @@ Returns the value of a vertex attribute on the vertex.
 | Parameter | Description | Data type
 
 | `attrName`
-| A vertex attribute
-| `STRING`
+| The name of an attribute. This can be a query parameter, constant string, or global string variable.
+| Base types: `STRING` `INT` `UINT` `DATETIME` `DOUBLE` `FLOAT` `BOOL`
+
+Container types: `LIST<INT/DATETIME/DOUBLE/STRING>`, `SET<INT/DATETIME/DOUBLE/STRING>`
```
yuchenZhangTG: dito

yuchenZhangTG (changes requested):

PR #83, opened 2022-11-22
update with notes on heapaccum features

alexthomasTG:


modules/querying/pages/accumulators.adoc
```diff
@@ -1251,6 +1251,12 @@ The sort key hierarchy is from left to right, with the leftmost key being the pr
 The keywords `ASC` and `DESC` indicate Ascending (lowest value first) or Descending (highest value first) sort order.
 Ascending order is the default.
 
+HeapAccum supports comparison with the `==` and the `!=` operators.
+If two HeapAccums hold the same types of tuple, and have the same sort order and direction, they are considered equivalent.
```
pengwangTigerGraph: @alexthomasTG Clarification: If two HeapAccums hold the same types of tuple, and have the same sort order and direction, they could be compared or assigned with each other, because they're deemed as the same kind/type of HeapAccum. I would prefer equivalent for the same kind of HeapAccums that hold the same tuples/no tuples.

PR #84, opened 2022-11-23
add note about loop variable restriction

alexthomasTG:


modules/querying/pages/func/query-user-defined-functions.adoc
```diff
@@ -67,6 +67,10 @@ You can write your functions in the `ExprFunctions` file provided as a sample in
 
 If your function requires a user-defined struct or helper function, that struct or helper function must be defined in a separate `ExprUtil` file.
 
+C{pp} supports changing the value of a `for` loop variable within a loop.
```
victorleeTG, line 71: Should we make this a NOTE?

victorleeTG (approved):
suggested but not required change

PR #85, opened 2022-11-23
To string types doc 1553

alexthomasTG:


quanchentg (commented):
type-conversion-functions.adoc looks good. No idea about what other two file's changes mean.

PR #93, opened 2023-02-10
update note about schema change best practices

alexthomasTG:


modules/ddl-and-loading/pages/modifying-a-graph-schema.adoc
```diff
@@ -20,13 +20,17 @@ A schema change will invalidate any loading jobs or query jobs which relate to a
 Invalid loading jobs are dropped, and invalid queries are uninstalled. After the schema update, the user will need to create and install new load and query jobs based on the new schema.
 ====
 
-Jobs and queries for unaltered parts of the schema will still be available and do not need to be reinstalled.  However, even though these jobs are valid (e.g., they *can* be run), the user may wish to examine whether they still perform the preferred operations (e.g., do you *want* to run them?)
+Jobs and queries for unaltered parts of the schema will still be available and do not need to be reinstalled.
+However, even though these jobs are valid (e.g., they *can* be run), consider whether they will still perform the desired operations after the change (e.g., do you *want* to run them?).
```
songting-chen: This is unclear. @junhyungshin-tg what do you think?

junhyungshin-tg: I think this part is unnecessary honestly.

> Jobs and queries for unaltered parts of the schema will still be available and do not need to be reinstalled.

First of all, users will expect their jobs/queries remain the same for sure, and the previous warning already implied that those for unaltered parts of the schema will unaffected. Also, there's no such installation concept for a job that could confuse users.

> However, even though these jobs are valid (e.g., they *can* be run), consider whether they will still perform the desired operations after the change (e.g., do you *want* to run them?).

Furthermore, we shouldn't suggest users to validate after they change schema that's something we need to guarantee. This sounds like the product is immature. This is also already a part of our integration test.

alexthomasTG: Good point, I'll remove this. Thanks!

songting-chen (approved):

PR #95, opened 2023-02-22
added token functions to list

alexthomasTG:


junhyungshin-tg (commented):
Just a minor comment about cosmetic issues in how each function signatures are written. For example, there's no margin spaces in `[find_in_set(_str_, _str_list_)]` (no spaces between an arg and parenthesis), but there are in `[soundex( str )]` and `space( n )`. Let's have a consistency check and we should be good to go as the contents here are majorly based on another page.

PR #140, opened 2024-06-26
[DOC-2113] feat(optimizer): Histogram Stats API Update and Enhancement;

romansod:
Makes several fixes to existing documentation while also adding all of the new implicit parameterization with examples

### DIFF
--- a/modules/gsql-v2/querying/pages/query-optimizer/stats-api.adoc
+++ b/modules/gsql-v2/querying/pages/query-optimizer/stats-api.adoc
@@ -21,13 +21,15 @@ include::partial$cardinality-api.adoc[]
 
 Histograms store information about the distribution of attribute values. They are computed using an Equi-Width computation method. This means, given a target bucket count, a histogram will be computed with up to that many buckets (see bucket parameter for more details). 
 
-**Buckets**: Increasing the number of buckets will increase the granularity of the computed histogram, while decreasing the number of buckets will increase compaction. Choosing the right bucket count should be done based on the use case first, then adjusted afterwards if necessary. Please see recommendations below for best practices.
+**Bucket Count**: Increasing the number of buckets will increase the granularity of the computed histogram, while decreasing the number of buckets will increase compaction. Choosing the right bucket count should be done based on the use case first, then adjusted afterwards if necessary. Please see recommendations below for best practices.
 
-NDV (Number of Distinct Values): this value is approximated for each bucket. For discrete datatypes (uint, int, datetime) this will represent the maximum potential NDV. For floating point, non discrete datatypes (float, double), assuming uniform distribution means assuming all values in a bucket are distinct. Primary id’s always have exact NDV values, since they are by definition all unique values.
+**NDV** (Number of Distinct Values): This value is approximated for each bucket. 
 
-Empty buckets in the histogram are denoted by a totalCount, upperCount, and ndv of zero. upperBound will have the max value for that bucket. This is only a theoretical limit for the bucket, and the totalCount should be relied on to determine whether a bucket is empty
+* Integral datatypes (UINT, INT, DATETIME): the NDV value represents an upper bound (the actual NDV may be lower).
+* Floating point numeric values (FLOAT, DOUBLE): the NDV is approximated by assuming all values are distinct. 
+* Primary id’s: the NDV represents the actual NDV and no approximation is used.
 
-Supported Data Types are: UINT, INT, DATETIME, FLOAT, DOUBLE
+**Supported Data Types**: UINT, INT, DATETIME, FLOAT, DOUBLE
 
 include::partial$histogram-api.adoc[]
 
--- a/modules/gsql-v2/querying/pages/query-optimizer/stats-api.adoc
+++ b/modules/gsql-v2/querying/pages/query-optimizer/stats-api.adoc
@@ -19,7 +19,15 @@ include::partial$cardinality-api.adoc[]
 
 == Histogram
 
-Histograms store information about the distribution of attribute values.
+Histograms store information about the distribution of attribute values. They are computed using an Equi-Width computation method. This means, given a target bucket count, a histogram will be computed with up to that many buckets (see bucket parameter for more details). 
+
+Buckets: Increasing the number of buckets will increase the granularity of the computed histogram, while decreasing the number of buckets will increase compaction. Choosing the right bucket count should be done based on the use case first, then adjusted afterwards if at all necessary. Please see recommendations below for best practices.
+
+NDV (Number of Distinct Values): this value is approximated for each bucket. For discrete datatypes (uint, int, datetime) this will represent the maximum potential NDV. For floating point, non discrete datatypes (float, double), assuming uniform distribution means assuming all values in a bucket are distinct. Primary id’s always have exact NDV values, since they are by definition all unique values.
+
+Empty buckets in the histogram are denoted by a totalCount, upperCount, and ndv of zero. upperBound will have the max value for that bucket. This is only a theoretical limit for the bucket, and the totalCount should be relied on to determine whether a bucket is empty
+
+Supported Data Types are: UINT, INT, DATETIME, FLOAT, DOUBLE
 
 include::partial$histogram-api.adoc[]
 
--- a/modules/gsql-v2/querying/pages/query-optimizer/stats-api.adoc
+++ b/modules/gsql-v2/querying/pages/query-optimizer/stats-api.adoc
@@ -21,7 +21,7 @@ include::partial$cardinality-api.adoc[]
 
 Histograms store information about the distribution of attribute values. They are computed using an Equi-Width computation method. This means, given a target bucket count, a histogram will be computed with up to that many buckets (see bucket parameter for more details). 
 
-Buckets: Increasing the number of buckets will increase the granularity of the computed histogram, while decreasing the number of buckets will increase compaction. Choosing the right bucket count should be done based on the use case first, then adjusted afterwards if at all necessary. Please see recommendations below for best practices.
+**Buckets**: Increasing the number of buckets will increase the granularity of the computed histogram, while decreasing the number of buckets will increase compaction. Choosing the right bucket count should be done based on the use case first, then adjusted afterwards if necessary. Please see recommendations below for best practices.
 
 NDV (Number of Distinct Values): this value is approximated for each bucket. For discrete datatypes (uint, int, datetime) this will represent the maximum potential NDV. For floating point, non discrete datatypes (float, double), assuming uniform distribution means assuming all values in a bucket are distinct. Primary id’s always have exact NDV values, since they are by definition all unique values.
 
--- a/modules/gsql-v2/querying/partials/histogram-api.adoc
+++ b/modules/gsql-v2/querying/partials/histogram-api.adoc
@@ -9,8 +9,8 @@ READ_SCHEMA privilege is required for the graph that the histogram attribute bel
 
 ==== Parameters
 Options can be specified in the header using -H:
-GSQL-Timeout indicates timeout for the request. In the case of multiple computations, this timeout covers the overall timeout, including all the histogram computations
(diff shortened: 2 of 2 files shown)


modules/gsql-v2/querying/pages/query-optimizer/stats-api.adoc
```diff
@@ -19,7 +19,15 @@ include::partial$cardinality-api.adoc[]
 
 == Histogram
 
-Histograms store information about the distribution of attribute values.
+Histograms store information about the distribution of attribute values. They are computed using an Equi-Width computation method. This means, given a target bucket count, a histogram will be computed with up to that many buckets (see bucket parameter for more details). 
+
+Buckets: Increasing the number of buckets will increase the granularity of the computed histogram, while decreasing the number of buckets will increase compaction. Choosing the right bucket count should be done based on the use case first, then adjusted afterwards if at all necessary. Please see recommendations below for best practices.
```
victorleeTG:
```suggestion
**Buckets**: Increasing the number of buckets will increase the granularity of the computed histogram, while decreasing the number of buckets will increase compaction. Choosing the right bucket count should be done based on the use case first, then adjusted afterwards if necessary. Please see recommendations below for best practices.
```

modules/gsql-v2/querying/pages/query-optimizer/stats-api.adoc
```diff
@@ -19,7 +19,15 @@ include::partial$cardinality-api.adoc[]
...
+
+NDV (Number of Distinct Values): this value is approximated for each bucket. For discrete datatypes (uint, int, datetime) this will represent the maximum potential NDV. For floating point, non discrete datatypes (float, double), assuming uniform distribution means assuming all values in a bucket are distinct. Primary id’s always have exact NDV values, since they are by definition all unique values.
```
victorleeTG:
```suggestion
**NDV** (Number of Distinct Values): This value is approximated for each bucket. For discrete datatypes (UINT, INT, DATETIME) this will represent the maximum potential NDV. For floating point, non discrete datatypes (float, double), assuming uniform distribution means assuming all values in a bucket are distinct. Primary ID’s always have exact NDV values, since they are by definition all unique values.
```

victorleeTG: I don't follow your explanation of NDV for different data types. I get the concept that the system knows you have X values in bucket B, but it doesn't know if the values are distinct. So, at the extreme end, all the values are distinct. What I don't get is the distinction you are making about floating point values. "non discrete datatypes"  - in a computer, floating point values are actually discrete. Maybe you are trying to say that true value 0.999999 and true value 1.000001 might both be represented by 1.000000. But I think the bottom line is that the histogram assumes all values are distinct. Is that the case or not?

modules/gsql-v2/querying/partials/histogram-api.adoc
```diff
@@ -3,11 +3,14 @@
...
+This endpoint computes histograms for a specified attribute of a vertex or edge type.
+
+READ_SCHEMA privilege is required for the graph that the histogram attribute belongs to. When gathering for multiple graphs at once (see Compute Multiple Histograms in one request), READ_SCHEMA is required on ALL of those graphs
 
 ==== Parameters
-No URL parameters.
-Provide parameters in the request body.
+Options can be specified in the header using -H:
+`GSQL-Timeout` indicates timeout for the request. In the case of multiple computations, this timeout covers the overall timeout, including all the histogram computations
+`GSQL-Thread-Limit` indicates the thread limit per histogram computation. In the case of multiple computations this thread limit will apply to each histogram computation individually
```
victorleeTG:
```suggestion
`GSQL-Timeout` indicates timeout for the request. In the case of multiple computations, this timeout covers the overall timeout, including all the histogram computations.
`GSQL-Thread-Limit` indicates the thread limit per histogram computation. In the case of multiple computations this thread limit will apply to each histogram computation individually.
```

modules/gsql-v2/querying/partials/histogram-api.adoc
```diff
@@ -67,10 +70,7 @@ Edge attribute::
 |`bucket`
 |The number of intervals (https://en.wikipedia.org/wiki/Data_binning[buckets]) to divide the range of attribute values by.
 
-It is recommended that each bucket have at least 100 values.
-Therefore, a good rule of thumb is to divide the number of vertices by 100 to decide the number of buckets.
-For example, if the vertex type has 100,000 vertices, set this parameter to no higher than 1000.
-A standard value for this parameter is 256.
+By default we use a bucket value of 256. With discrete datatypes, a bucket count that is lower than the specified target may be necessary depending on the data the histogram is computed over.
```
victorleeTG: What do you mean by "specified target"? 
Did you mean "default value"?

modules/gsql-v2/querying/partials/histogram-api.adoc
```diff
@@ -84,31 +84,166 @@ A standard value for this parameter is 256.
...
+  ]
+}
+----
+--
+====
+
+==== Compute Multiple Histograms together
 
-To compute the histogram data for attribute `creationDate`  from vertex type `Person` to vertex type `Comment`, make the following request:
+In addition to individually requesting histogram computation through the POST endpoint, it is also possible to gather ALL histograms on a per schema object basis. `bucket` parameter can still be specified and will be applied to every histogram being computed in the request
```
victorleeTG:
```suggestion
In addition to requesting individual histograms through the POST endpoint, it is also possible to request all valid histograms on a per schema object basis.  The`bucket` parameter can still be specified and will be applied to every histogram being computed in the request.
```

modules/gsql-v2/querying/partials/histogram-api.adoc
```diff
@@ -84,31 +84,166 @@ A standard value for this parameter is 256.
...
+// (default bucket value)
+-d '{"graph":"test_graph", "edge":"COMP_MEMBER", "attribute":"joinDate"}'
+// Compute Histograms for ALL valid attributes under:
+// graph test_graph
+// edge COMP_MEMBER, from vertex compForum, to vertex compPerson
+// (default bucket value)
+-d '{"graph":"test_graph", "edge":"COMP_MEMBER", "from":"compForum", "to":"compPerson"}'
+----
+
+The output generated is in the same format as individual computation, with the `success` field tracking the total number of successful computations and `failure` will track all of the failures encountered:
```
victorleeTG:
```suggestion
The output generated is in the same format as individual computation, with the `success` field tracking the total number of successful computations and `failure` tracking all of the failures encountered:
```

modules/gsql-v2/querying/partials/histogram-api.adoc
```diff
@@ -84,31 +84,166 @@ A standard value for this parameter is 256.
...
+}
+----
+--
+====
+
+A more condensed error message will be provided under `failure` if any of the histogram computations encounter an error. This allows the request to log the error and continue with the next computation. For more information on the specific error refer to the logs or run the particular histogram computation individually for more verbose feedback. 
 
 === Retrieve histogram statistics
 
+`READ_DATA` privilege is required to retrieve a histogram for a particular attribute. This can be at the attribute level or a higher level. Note that for edge based histograms, the 'from' and 'to' vertex are not read and therefor only the edge requires the privilege
```
victorleeTG:
```suggestion
`READ_DATA` privilege is required to retrieve a histogram for a particular attribute. This can be at the attribute level or a higher level. Note that for edge-based histograms, the 'from' and 'to' vertex are not read and therefore only the edge requires the privilege.
```

modules/gsql-v2/querying/partials/histogram-api.adoc
```diff
@@ -196,16 +533,87 @@ This includes histogram data on all vertex attributes in that graph.
...
+// vertex Comment
+-d '{"graph":"test_graph", "vertex":"Comment"}'
+// Delete Histograms for ALL attributes under:
+// graph test_graph
+// edge COMP_MEMBER, from vertex compForum, to vertex compPerson
+-d '{"graph":"test_graph", "edge":"COMP_MEMBER", "from":"compForum", "to":"compPerson"}'
+----
+
+==== Current Limitations
+If leader switch or some other interruption takes place in GSQL we will not be able to resume progress. Statistics are stored as they are computed along with their start and end timestamps so there is no progress lost. See recommendations for some suggestions on how to handle such situations
```
victorleeTG:
```suggestion
If leader switch or some other interruption takes place in GSQL the histogram operation will not be able to resume. Statistics are stored as they are computed along with their start and end timestamps so there is no progress lost. See recommendations for some suggestions on how to handle such situations.
```

modules/gsql-v2/querying/partials/histogram-api.adoc
```diff
@@ -196,16 +533,87 @@ This includes histogram data on all vertex attributes in that graph.
...
+
+==== Recommendations
+1. If timeout or some other event interrupts the collection of all the desired statistics, specifying a smaller subset of histograms based on what still has yet to be gathered will reduce the amount of regathering. For example, if executing a histogram computation request for attributes under all graphs, and after 2/4 graphs are fully gathered we are unable to continue, instead of regathering on all graphs again we can issue two requests at the graph level for graphs 3 and 4
```
victorleeTG:
```suggestion
1. If timeout or some other event interrupts the collection of all the desired statistics, specifying a smaller subset of histograms based on what still has yet to be gathered will reduce the amount of regathering.
For example, if executing a histogram computation request for attributes under all graphs, and after 2/4 graphs are fully gathered a timeout occurs, you can issue two requests at the graph level for graphs 3 and 4.
```

modules/gsql-v2/querying/partials/histogram-api.adoc
```diff
@@ -196,16 +533,87 @@ This includes histogram data on all vertex attributes in that graph.
...
+
+2. When possible, it is best to schedule histogram computations after schema change jobs and data loading, not before. This is to ensure the statistics best reflect the current state of the database
+
+3. When choosing the right bucket value, the first step is to determine the requirements of the use case. Then, consider the granularity and compaction desired based on accuracy and performance / storage consumption
```
victorleeTG:
```suggestion
2. When possible, it is best to schedule histogram computations after schema change jobs and data loading, not before. This is to ensure the statistics best reflect the current state of the database.

3. When choosing the right bucket value, the first step is to determine the requirements of the use case. Then, choose the fewer number of buckets that provides enough granularity.
```

victorleeTG (changes requested):
A combination of
1) Statements that aren't clear enough.
2) Missing periods at ends of sentences (maybe run a spell checker?)
3) Style issues: avoid "we" when you mean the "the product"; capitalization of GSQL keywords; consistent verb tense; ...

modules/gsql-v2/querying/pages/query-optimizer/stats-api.adoc
romansod: I see your point. Your understanding of my intention is correct. Maybe it is easier if I refer to integral vs floating point?

The distinction here is as follows:

1. Primary_ID regardless of data type will always have exact ndv
2. Integral (uint, int, datetime) approximate ndv with a max ndv (ie the actual ndv may be lower but never higher)
3. Floating point (float, double) approximate ndv by assuming all values are distinct

modules/gsql-v2/querying/partials/histogram-api.adoc
romansod: By specified target, I am trying to explain how, if we are using say an integer data type with the following values:
```
0, 1, 2, 3
```
If the user suggests a target bucket number of say 10, we will instead use 4. So the user specified target bucket count is a target bucket count, but we may reduce that if it is simply not compatible to the maximum, sensible bucket number (in the example, 4)

modules/gsql-v2/querying/pages/query-optimizer/stats-api.adoc
victorleeTG: yes, integral would be better than discrete.

victorleeTG (approved):

PR #143, opened 2024-07-12
[DOC-2228] fix(OpenCypher): Remove PercentileCont from the document;

DavidFan2000:
JIRA Link: https://graphsql.atlassian.net/browse/DOC-2228

- Looks like PercentileCont was added to the document by mistake. We should remove this from our 3.10 document.


victorleeTG (approved):
Well done

PR #161, opened 2024-07-30
[DOC-2075] docs(gsql): create the documentation of packaged template queries and FilterType;

xuanleilin:
Ticket Link: https://graphsql.atlassian.net/browse/DOC-2075.

### DIFF
--- a/modules/template_queries/nav.adoc
+++ b/modules/template_queries/nav.adoc
@@ -1,0 +1,3 @@
+* xref:index.adoc[Template Queries]
+** xref:user-defined-template-queries.adoc[User-defined Template Queries]
+** xref:filtertype.adoc[FilterType]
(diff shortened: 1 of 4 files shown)


arun-TG-PM (commented):
Try not to use direct company name such as BOA. Use generic examples such as customer A , company A etc

arun-TG-PM (approved):

victorleeTG:
@xuanleilin  **Writing** template queries and adding them to a package is not yet a user-facing feature. 
So far, it is only used for algorithms that TigerGraph wrote.  Yesterday (Aug 27) we discovered a bug in the template string replacement.

However, this is good internal information which should have been shared with the the ML team because they have been trying to guess at how the string translation works.

PR #165, opened 2024-07-31
[DOC-2248] feat(OpenCypher): update the list of supported & unsupported functions (4.1)

DavidFan2000:
Jira Link: https://graphsql.atlassian.net/browse/DOC-2248
- Organize the list of supported OpenCypher functions so everything is in Alphabetical order
- Add the function elementId to supported functions
- Move the following functions from Unsupported Functions to Supported Functions:
  - head
  - id
  - last
  - range
  - size (on list)
  - split
  - tail
  - stdev
  - stdevp
- Move the following functions from Supported Functions to Unsupported functions (because they were partially supported, or not correctly supported):
  - coalesce
  - type


.gitignore
```diff
@@ -1,2 +1,3 @@
 .DS_Store
-.idea
```
DavidFan2000, line 2: From the gitignore, it looks like we always wanted to ignore the `.idea` folder.
This folder can only be correctly ignored after we remove it from git. Otherwise git will keep tracking the changes.

modules/openCypher-in-gsql/pages/openCypher-in-gsql.adoc
```diff
@@ -107,19 +107,7 @@ See below for restrictions.
 
 [cols="1,1"]
 |===
-|left() |String Returns a string containing the specified number of leftmost characters of the original string.
```
DavidFan2000, line 110: Moving them around to keep everything in alphabetical order.

modules/openCypher-in-gsql/pages/openCypher-in-gsql.adoc
```diff
@@ -129,67 +117,93 @@ See below for restrictions.
 
 |atan2() |Trigonometric Returns the arctangent2 of a set of coordinates in radians.
 
+|avg() |Aggregating Returns the average of a set of values.
```
qianzhang-tg: We should unify the description of these functions with latest `openCypher` language reference's language. The latest I can find is this [ref](https://s3.amazonaws.com/artifacts.opencypher.org/openCypher9.pdf) 
For example: `avg()`'s description is `returns the average of a set of numeric values` in above reference.

If you have newer version of openCypher language reference, then use that newest.

DavidFan2000: Great catch! 

I moved `avg` from the section of unsupported functions to section of supported functions, but didn't cross check it with OpenCypher document. Just updated the description to include the updated description.

DavidFan2000: I think in the future we may also want to group the OpenCypher functions (for example, group aggregating functions, scalar functions, and others into different sections). Though I would do that later because I think other people are also working on improving the overall formatting.

Tushar-TG-14 (changes requested):
@victorleeTG 
I’ve reviewed the pull request, and the updates are correct:

- The elementId function has been added to supported functions.
- Functions have been accurately moved from Unsupported to Supported, and vice versa.

Please take a look to confirm before merging.

Tushar-TG-14 (approved):

PR #179, opened 2024-10-15
DOC-2028-missing-string-funcs

victorleeTG:


modules/querying/pages/func/mathematical-functions.adoc
```diff
@@ -477,17 +503,30 @@ Returns the square root of a number
...
+The `decimal_places` parameter can be negative, meaning the target precision is to the left of the decimal point.
+This effectively rounds the number to some power of 10.
 
 
 ==== Return type
 
-A numeric type.
+A numeric type:
+
+* The same type as the input, if `decimal places` is positive.
```
DavidFan2000: In the current implementation, our type checker records the `trunc` function as having type `double -> double` or `(double, int64) -> double`. The return type is technically double, but it will be compatible with any numeric type due to automatic type conversion. I'm not sure how much information we want to provide here.

modules/querying/pages/func/mathematical-functions.adoc
```diff
@@ -500,19 +539,19 @@ A numeric type.
...
 |===
 
 
 ==== Examples
 
-[,text]
+[source,text]
 ----
-trunc(9.99) => 9
-trunc(-9.99) => 9
```
DavidFan2000, line 513: This was a typo here. We can add another example to show how negative numbers are still negative after truncation. (`trunc(-9.9) => -9`)

modules/querying/pages/func/mathematical-functions.adoc
victorleeTG: @DavidFan2000 Thanks. I'll just say the return type is double.

PR #200, opened 2024-11-26
DOC-1944

Tushar-TG-14:


modules/querying/pages/func/type-conversion-functions.adoc
```diff
@@ -189,4 +189,116 @@ Converts a number to a string using the C++ printf function with the `g` specifi
...
+
+`toBoolean(str)`
+
+  
+===  Description
+
+This function converts a string into a boolean value (`true` or `false`).
+
+* If the input is already a boolean, it returns the same value.
+* If the input is a string, the function tries to parse it into a boolean. For example, "true" becomes true, and "false" becomes false.
```
victorleeTG: "Tries to parse" is vague because it doesn't explain what type of inputs can be parsed or not.
I think it would be better to say:

- If the string is the string equivalent of boolean `true` or `false` (case insensitive), it returns the corresponding boolean value.
- Otherwise it returns NULL.

modules/querying/pages/func/type-conversion-functions.adoc
```diff
@@ -189,4 +189,116 @@ Converts a number to a string using the C++ printf function with the `g` specifi
...
-----
\ No newline at end of file
+----
+
+== `toBoolean()`
+
+  
+===  Syntax
+
+`toBoolean(str)`
```
victorleeTG: The input can be either a string or a boolean, so it is not correct to show the input type as `str`.  
Compare to Neo4j's documentation.

modules/querying/pages/func/type-conversion-functions.adoc
```diff
@@ -189,4 +189,116 @@ Converts a number to a string using the C++ printf function with the `g` specifi
...
+toBoolean("false") -> false  
+toBoolean("invalid") -> null
+----
+
+== `toFloat()`
+
+  
+===  Syntax
+
+`toFloat(num)`
```
victorleeTG: The input type is not limited to `num`.

modules/querying/pages/func/type-conversion-functions.adoc
```diff
@@ -189,4 +189,116 @@ Converts a number to a string using the C++ printf function with the `g` specifi
...
+
+  
+===  Description
+
+Converts an integer or string to a floating-point number.
+
+* If the input is already a floating-point number, the function will return it unchanged.
+* If the input is a valid number in string format (e.g., "14.5"), the function converts it to a float (e.g., 14.5).
```
victorleeTG: Is scientific notation supported, such as "1.45E-6" ?

modules/querying/pages/func/type-conversion-functions.adoc
```diff
@@ -189,4 +189,116 @@ Converts a number to a string using the C++ printf function with the `g` specifi
...
+toFloat(14) -> 14.0  
+toFloat("-3.5") -> -3.5  
+----
+
+== `toInteger()`
+
+  
+===  Syntax
+
+`toInteger(num)`
```
victorleeTG: Same comment as above.

modules/querying/pages/func/type-conversion-functions.adoc
```diff
@@ -189,4 +189,116 @@ Converts a number to a string using the C++ printf function with the `g` specifi
...
+| `num`
+| The number to convert into an integer
+| Number
+|===
+
+=== Example
+
+----
+toInteger(14.7) -> 14  
+toInteger("-10") -> -10
```
victorleeTG:
```suggestion
toInteger("-10") -> -10
toInteger("five") -> NULL
```
It's good to include an example of an input that a user might think is valid, but isn't.

victorleeTG (changes requested):

victorleeTG (approved):

PR #207, opened 2024-12-13
DOC-2333

anjaliIthapeTG:


modules/querying/pages/modules/querying/pages/write-query-data-to-cloud.adoc
```diff
@@ -0,0 +1,57 @@
...
+* These parameters are set per session, so each user will need to provide their own credentials if they’re working with S3.
+* All queries run during this session will use these credentials to write to S3.
+
+[NOTE]:If both methods (using gadmin config and GSQL session parameters) are applied simultaneously, the GSQL session parameters take precedence.
+
+When you execute queries in TigerGraph that produce results, the query output can be written directly to the configured S3 bucket. Since S3 is a shared storage system, multiple nodes in a TigerGraph cluster might try to upload results to the same bucket at the same time.
+
+== Unique File Paths to Avoid Conflicts
+
+To avoid naming conflicts (multiple nodes trying to write to the same file), TigerGraph allows you to include a unique prefix in the file path. The path can include:
```
victorleeTG: "allows you to"
Is this an option that the user needs to make use of?  Does the user get to choose the suffix format?  It's not clear.  If the suffix format is fixed (predetermined), then don't say "allows you to".

modules/querying/pages/modules/querying/pages/write-query-data-to-cloud.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+* Instance Name: A suffix like \GPE_{PartitionId}_{ReplicaId} ensures uniqueness by identifying the instance that generated the output.
+Distributed Queries: For distributed queries, TigerGraph appends suffixes to differentiate between the roles of nodes in the cluster. The two roles are:
+* Coordinator: The node managing the query (written as _coordinator).
+Worker: The node processing the query (written as _worker).
```
victorleeTG:
```suggestion
* Role: For distributed queries, TigerGraph appends suffixes to differentiate between the roles of nodes in the cluster. The two roles are:
** Coordinator: The node managing the query (written as _coordinator).
** Worker: The node processing the query (written as _worker).
```
Fixing the organization and indentation of the information.

modules/querying/pages/modules/querying/pages/write-query-data-to-cloud.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+So, your S3 file paths might look like:
+
+* \GPE_{PartitionId}_{ReplicaId}_coordinator
+* \GPE_{PartitionId}_{ReplicaId}_worker
+
+[NOTE]: The output is saved with the format “File_name + Suffixes”.
```
victorleeTG: Rather than a note, I think an example would be more appropriate. Something like,

For example, if a 3 x 2 TigerGraph cluster is running a distributed query and sending the output to a file "myResultsPath", then there may actually be 3 output files:
* myResultsPath\GPE_1_0_worker
* myResultsPath\GPE_1_1_coordinator
* myResultsPath\GPE_1_2_worker

This example is my **guess** at how this feature works, based on reading your documentation.

victorleeTG (changes requested):

modules/querying/pages/modules/querying/pages/write-query-data-to-cloud.adoc
```diff
@@ -0,0 +1,64 @@
...
+SET s3_aws_secret_access_key = <YOUR_AWS_SECRET_ACCESS_KEY>
+----
+* These parameters are set per session, so each user will need to provide their own credentials if they’re working with S3.
+* All queries run during this session will use these credentials to write to S3.
+
+[NOTE]:If both methods (using gadmin config and GSQL session parameters) are applied simultaneously, the GSQL session parameters take precedence.
+
+== Unique File Paths to Avoid Conflicts
+
+Since S3 is a shared storage system, multiple nodes in a cluster can upload to the same S3 bucket.To avoid naming conflicts (multiple nodes trying to write to the same file),the S3 path can include a prefix based on the instance name :
```
victorleeTG:
```suggestion
Since S3 is a shared storage system, multiple nodes in a cluster can upload to the same S3 bucket.
To avoid naming conflicts (multiple nodes trying to write to the same file), the S3 path can include a prefix based on the instance name:
```

victorleeTG: I see you changed the verb to "can". Better, but I still have some uncertainty about how this feature works. Let's discuss.

anjaliIthapeTG: Added "will" for better understanding

victorleeTG (approved):

PR #217, opened 2025-01-17
DOC-2374

anjaliIthapeTG:


modules/querying/pages/query-operations.adoc
```diff
@@ -571,6 +571,8 @@ Required privilege: No specific query privilege required.
 This syntax is similar in concept to SQL queries.
 Queries are not named, do not accept parameters, and are not saved after being run.
 
+For [runOptions], refer to the syntax in the section xref:gsql-docs:querying:query-operations.adoc#_run_query[RUN QUERY].
```
victorleeTG, line 574:
```suggestion
For [runOptions], refer to the syntax in the section xref:query-operations.adoc#_run_query[RUN QUERY].
```
REMEMBER THIS RULE: In a xref, use the MOST LOCAL path that you can.  In particular, NEVER mention the name of the component if the link's source and destination are in the same component. I have explain this rule before and what happens if you break the rule: the link will jump to the most current version, regardless of what version the user had started from.

modules/querying/pages/query-operations.adoc
```diff
@@ -705,7 +707,7 @@ You can also run an installed query through REST requests - see xref:tigergraph-
 runQuery := (RUN | INTERPRET) QUERY [runOptions] queryName
     "(" parameterValueList | parameterValueJSON ")"
 
-runOptions := ( "-av" | "-d" )*
+runOptions := ( "-async"| "-av" | "-queue" <Workload Queue Name>)*
```
victorleeTG, line 710: @anjaliIthapeTG You can't just drop a feature or option (-d) as a side effect.  Think of the users. Please ask the developers when this option was dropped. Then think about how we should address this in documentation.

modules/querying/pages/query-operations.adoc
```diff
@@ -925,6 +927,14 @@ $ curl -s -H "GSQL-ASYNC:true" GET "http://localhost:14240/restpp/query/<graphNa
 
 To check the status and results of the queries executed in Detached Mode, use the xref:tigergraph-server:API:built-in-endpoints.adoc#_check_query_status_detached_mode[*`/query_status`*] and the xref:tigergraph-server:API:built-in-endpoints.adoc#_check_query_results_detached_mode[*`/query_result`*] RESTPP endpoints.
 
+=== -queue option
+
+When running a query, you can specify the workload queue to run the query on.
+If the queue is not specified, the query will be routed to the default queue.
+To specify the queue in the GSQL shell, you can use the `-queue` option, e.g.
```
victorleeTG, line 934:
```suggestion
For example, to have query `q1` run in the `AdHoc` queue:
```
COMMENT: 
1. "can" is the wrong verb here.
2. You could also show the syntax for running it as a REST endpoint, as we see in the examples for the -async option.

modules/querying/pages/query-operations.adoc
```diff
@@ -560,7 +560,7 @@ There are two GSQL syntax options for Interpreted GSQL:
 .Syntax for interpreting an anonymous query
 [source,ebnf]
 ----
-interpretQuery := INTERPRET QUERY "(" ")"
+interpretQuery := INTERPRET QUERY [runOptions] "(" ")"
```
victorleeTG, line 563: According to Dadong's recent comment, `runOptions` are not supported here.

modules/querying/pages/query-operations.adoc
victorleeTG: According to Dadong's recent comment, -d is still functional, but there is a better feature -PROFILE. Is that in 4.2?  Let's not add a description of -d.

victorleeTG (changes requested):

PR #219, opened 2025-01-20
DOC-2078

Tushar-TG-14:


modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,101 @@ The scope of the command depends on the user's current scope.
 If the user has set a working graph, then `DROP ALL` removes all the queries for that graph.
 If a user has set their scope to be global, then `DROP ALL` removes all queries across all graph spaces.
 
+== Query Descriptions
+
+Query descriptions explains the procedures for updating, showing, and dropping descriptions of queries and query parameters.
```
victorleeTG:
```suggestion
Query descriptions explain the procedures for updating, showing, and dropping descriptions of queries and query parameters.
```

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,101 @@ The scope of the command depends on the user's current scope.
...
+----
+
+Required privilege
+`WRITE_QUERY`
+
+To update a description, run the `UPDATE DESCRIPTION` command as shown below. You can update the description of a query or a query parameter. If you choose not to specify a graph in the command, the current scope will be used as the scope of the description.
+
+[source, gsql]
+----
+GSQL > USE GRAPH POC_GRAPH
```
victorleeTG: In other places in our documentation, "POC_GRAPH" is lower case: "poc_graph".
If this is just the name of an example graph, it's clearer if it's not all capital letters, because we want the reader to see the difference between what is a keyword and what is not.

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,101 @@ The scope of the command depends on the user's current scope.
...
+GSQL > UPDATE DESCRIPTION OF QUERY query1 "xxx"
+GSQL > UPDATE DESCRIPTION OF QUERY query2 ON GRAPH graph1 "xxx"
+GSQL > UPDATE DESCRIPTION OF QUERY_PARAM query1.param1 "xxx"
```
victorleeTG: If these are supposed to be examples, why not put more realistic examples instead of "xxx"?
In particular, each of these 3 should have a different description.

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,101 @@ The scope of the command depends on the user's current scope.
...
+====
+The term `UPDATE` here means adding a description to the target.
+If the target does not already have a description, it will be added.
+If the target already has a description, it will be replaced with the new one.
+====
+
+[CAUTION]
+====
+Using create or replace is preferable for updating a query to ensure that the description is retained.
+Otherwise, there is a risk of losing the description.
```
victorleeTG: I was a briefly confused by this, because the actual meaning here is `CREATE OR REPLACE QUERY`, not just the regular English works "create or replace".  Please write it as code. (Maybe you were just copying what the developer wrote and didn't realize this is code?)

Also, I think it's clearer to a reader if the sentence starts by established the context:

When updating a query, it is recommended to use `CREATE OR REPLACE QUERY` instead of simply `CREATE QUERY` to ensure the description is retained.

By the way, this warning worries/confused me. Is the developer saying the product is unreliable?

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,101 @@ The scope of the command depends on the user's current scope.
...
+`READ_QUERY`
+
+To display a description, use the `SHOW DESCRIPTION` command as shown below. Similar to the update command, you can display the description of a query or a query parameter.
+
+
+[source, gsql]
+----
+GSQL > USE GRAPH POC_GRAPH
+GSQL > SHOW DESCRIPTION OF QUERY query1
+GSQL > SHOW DESCRIPTION OF QUERY query1,query2
```
victorleeTG: This example shows that is is possible to have a comma-separated list of queries or parameters. The syntax above does not indicate that, so technically, the syntax description is wrong, or at least overly simplified.

A simply way to do this would be to replace <objectName> with <objectName(s)>

and below the syntax, add
"where `<objectName(s)>` can be a comma-separated list of names.

We can do this sort of natural language commentary with this "informal" syntax. We could not do it with EBNF, which has very strict rules. ENBF is supposed to be understood by a computer program, e.g. a syntax checker could read your EBNF rules and your code, and then tell you if you code follows the EBNF rules.

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,101 @@ The scope of the command depends on the user's current scope.
...
+----
+
+Required privilege
+`WRITE_QUERY`
+
+To drop a description, run the `DROP DESCRIPTION` command like below.
+Similar to the update command, you can drop the description of a global type object or a local type object.
+
+To remove a description, use the DROP DESCRIPTION command as shown below.
+Similar to the update command, you can remove the description of a query or a query parameter.
```
victorleeTG: This seems repetitive/redundant.  Is the second paragraph trying to say something new that the first paragraph didn't already say?

Tip: In regular English, the base meaning of "drop" is to fall, but in databases it has a particular technical interpretation.  Whereas "remove" means to take away. We want to use very straightforward, non-metaphorical English in our explanations. Therefore, I would use "remove" to explain what `DROP` means.

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,101 @@ The scope of the command depends on the user's current scope.
...
+
+[source, gsql]
+----
+GSQL > USE GRAPH POC_GRAPH
+GSQL > DROP DESCRIPTION OF QUERY query1
+GSQL > DROP DESCRIPTION OF QUERY query1*,query2*
+GSQL > DROP DESCRIPTION OF QUERY * ON GRAPH poc_graph
+GSQL > DROP DESCRIPTION OF QUERY_PARAM query1.param1
+GSQL > DROP DESCRIPTION OF QUERY_PARAM query1.param1,query1.param2
+GSQL > DROP DESCRIPTION OF QUERY_PARAM query1*
```
victorleeTG: This example has `query1*` whereas the one above has `query2.*`:
one include a dot before the `*`, and one does not.  Are both supported, or is this is mistake?

Overall, I had a lot more comments on this than I thought I would. Are you read this as carefully as though you are trying to learn this well enough so that you could confidently write correct code?

victorleeTG (changes requested):

modules/querying/pages/query-operations.adoc
Tushar-TG-14: @victorleeTG 
This is what engineer has suggested:

The context is that when we use `CREATE QUERY` in GraphStudio, if the query already exist, it will first drop the query(which will drop the query's description together) then recreate the query. So this will losing all the descriptions about the query. It's better to use `CREATE OR REPLACE QUERY` then it will not drop the query.

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,99 @@ The scope of the command depends on the user's current scope.
 If the user has set a working graph, then `DROP ALL` removes all the queries for that graph.
 If a user has set their scope to be global, then `DROP ALL` removes all queries across all graph spaces.
 
+== Query Descriptions
+
+Query descriptions explain the procedures for updating, showing, and dropping descriptions of queries and query parameters.
```
victorleeTG: I think you mixed up two separate things into one sentence:
1) What is a query description?
2) What is contained in this section of the documentation?
It would nice to have one concise sentence for each of these.

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,99 @@ The scope of the command depends on the user's current scope.
...
+----
+
+Support is provided for displaying descriptions of multiple targets simultaneously. You can separate the targets using commas, or use the wildcard `*` to display all descriptions.
+
+=== Drop description
+
+*Syntax*
+[source]
+----
+DROP DESCRIPTION OF <objectType> <objectName> [ON GRAPH <graphName>]
```
victorleeTG:
```suggestion
DROP DESCRIPTION OF <objectType> <objectName(s)> [ON GRAPH <graphName>]
```

victorleeTG (changes requested):

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,99 @@ The scope of the command depends on the user's current scope.
...
+Required privilege
+`WRITE_QUERY`
+
+To remove a description, run the `DROP DESCRIPTION` command as shown below. This command can be used to remove the description of both global and local type objects, as well as queries and query parameters.
+
+[source, gsql]
+----
+GSQL > USE GRAPH poc_graph
+GSQL > DROP DESCRIPTION OF QUERY query1
+GSQL > DROP DESCRIPTION OF QUERY query1*,query2*
```
victorleeTG: `GSQL > DROP DESCRIPTION OF QUERY query1*,query2*`

I think we have another question for a developer. 
What does `query1*` mean?  It seems to say "a name that begins with `query1` and then has any other characters. For example, `query1A`, `query1_is_the_best` and `query1_is_my_favorite` would fit this pattern.

In all the other examples, `*` could only be used to represent an entire name, not part of a name. My guess is that either `*` actually cannot be used for part of a name OR it ALWAYS can be used for the suffix of a name. We need the developer to clarify

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,99 @@ The scope of the command depends on the user's current scope.
...
+GSQL > DROP DESCRIPTION OF QUERY * ON GRAPH poc_graph
+GSQL > DROP DESCRIPTION OF QUERY_PARAM query1.param1
+GSQL > DROP DESCRIPTION OF QUERY_PARAM query1.param1,query1.param2
```
victorleeTG, line 1239: While we are talking to the developer, in  SHOW DESCRIPTION QUERY_PARAM and DROP DESCRIPTION QUERY_PARAM,
if there is a list of multiple parameters, can they come from different queries, e.g.
`query1.param1, query2.param2`?

Tushar-TG-14:
@victorleeTG 
I’ve received the answers from Adil:

1. For `DROP DESCRIPTION OF QUERY query1, query2`, what does query1* mean? 
- Does it include names starting with `query1` (e.g., `query1A`, `query1_example`), or is `*` only for representing entire names? 
- Can `*` be used as a suffix or in other parts of a name, or does it have specific limitations?**

Answer:
The `*` in the `DROP DESCRIPTION` command represents all descriptions of queries or query parameters. 
I believe this was a typo. In general, the grammar for the DROP DESCRIPTION command allows users to specify names in the following ways:
- `identifier`: Drops the description of a query.
- `identifier.*`: Drops descriptions of all parameters for a query.
- `*`: Drops descriptions of all queries or query parameters across a graph.

Alternatively, users can write:
`DROP DESCRIPTION OF QUERY *` to drop the descriptions of all queries on a graph
`DROP DESCRIPTION OF QUERY_PARAM *` to drop the descriptions of all query parameters on a graph

Therefore the examples for the `DROP` command should be:
```
GSQL > USE GRAPH POC_GRAPH
GSQL > DROP DESCRIPTION OF QUERY query1
GSQL > DROP DESCRIPTION OF QUERY * ON GRAPH poc_graph
GSQL > DROP DESCRIPTION OF QUERY ALL ON GRAPH poc_graph
GSQL > DROP DESCRIPTION OF QUERY_PARAM query1.param1
GSQL > DROP DESCRIPTION OF QUERY_PARAM query1.param1,query1.param2
GSQL > DROP DESCRIPTION OF QUERY_PARAM query1.*
GSQL > DROP DESCRIPTION OF QUERY_PARAM * ON GRAPH poc_graph
GSQL > DROP DESCRIPTION OF QUERY_PARAM ALL ON GRAPH poc_graph
```

2. For `SHOW DESCRIPTION QUERY_PARAM` and `DROP DESCRIPTION QUERY_PARAM`, can parameters from different queries be combined in one command?
- For example, is `query1.param1,query2.param2` valid syntax?

Answer:
`SHOW DESCRIPTION`: You cannot combine parameters from different queries in a single command (no comma-separated list supported).
`DROP DESCRIPTION`: You can combine parameters from different queries in a single command.

Looking at the grammar source code, here are the differences between the `SHOW DESCRIPTION` command and the `DROP DESCRIPTION` command:
- `SHOW` does not support a comma-separated list of names, whereas `DROP` does.
- `SHOW` supports a wider range of regular expressions, where as `DROP` does not.

Therefore the examples for the `SHOW` command should be:
```
GSQL > USE GRAPH POC_GRAPH
GSQL > SHOW DESCRIPTION OF QUERY query1
GSQL > SHOW DESCRIPTION OF QUERY query1*
GSQL > SHOW DESCRIPTION OF QUERY * ON GRAPH graph1
GSQL > SHOW DESCRIPTION OF QUERY_PARAM query1.param1
GSQL > SHOW DESCRIPTION OF QUERY_PARAM query1.param2
GSQL > SHOW DESCRIPTION OF QUERY_PARAM query2.*
```

3. How does the wildcard `*` work with query parameters?
- Does `query1.*` include all parameters under `query1`?
- Can `*` be used across multiple queries, like `query1.*,query2.*`?

Answer:
- Yes, `query1.*` includes all parameters under `query1`.
- Yes, you can use the wildcard across multiple queries, like `query1.*, query2.*`.

4. Are there any limitations on using `*` with `SHOW` or `DROP` DESCRIPTION commands that we should document?

Answer:
No specific limitations.

Tushar-TG-14:
@victorleeTG 
I’ve received the answers from the engineer:

1. Can we remove `ALL` from the `DROP` command example?

Answer:
Yes, we can remove `ALL` from the DROP command example. The `DROP` command supports both `*` and `ALL`, but for consistency, we can remove the `ALL` part.

`DROP DESCRIPTION OF QUERY_PARAM *` and `DROP DESCRIPTION OF QUERY_PARAM ALL` is not supported. Must specify a query name. `{query_name}.{param_name}`

2. `SHOW` command example has `query1*` while `DROP` doesn't. Why?

Answer:
The `SHOW` command supports `query1*` to match queries with names starting with "query1", while `DROP` doesn't. This is because `SHOW` supports a wider range of regular expressions, whereas `DROP` does not.

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,104 @@ The scope of the command depends on the user's current scope.
...
+GSQL > SHOW DESCRIPTION OF QUERY query1
+GSQL > SHOW DESCRIPTION OF QUERY query1*
+GSQL > SHOW DESCRIPTION OF QUERY * ON GRAPH poc_graph
+GSQL > SHOW DESCRIPTION OF QUERY_PARAM query1.param1
+GSQL > SHOW DESCRIPTION OF QUERY_PARAM query2.*
+----
+
+* For the `SHOW DESCRIPTION` command, you can display the descriptions of individual queries or query parameters. Each query or parameter must be specified separately, and you cannot combine them in a single request.
+* The wildcard `*` can be used to show descriptions of all queries or all parameters for a specific query. 
+* `SHOW DESCRIPTION` also supports a wider range of regular expressions, allowing for more advanced querying when searching for descriptions.
```
victorleeTG:
```suggestion
```
Again, this was an answer to a specific question that we asked Adil. It will make no sense to a reader:
"A wider range of regular expressions" --> wider than what? You didn't say.  Why is that even important to know?

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,104 @@ The scope of the command depends on the user's current scope.
...
+----
+GSQL > USE GRAPH poc_graph
+GSQL > SHOW DESCRIPTION OF QUERY query1
+GSQL > SHOW DESCRIPTION OF QUERY query1*
+GSQL > SHOW DESCRIPTION OF QUERY * ON GRAPH poc_graph
+GSQL > SHOW DESCRIPTION OF QUERY_PARAM query1.param1
+GSQL > SHOW DESCRIPTION OF QUERY_PARAM query2.*
+----
+
+* For the `SHOW DESCRIPTION` command, you can display the descriptions of individual queries or query parameters. Each query or parameter must be specified separately, and you cannot combine them in a single request.
```
victorleeTG:
```suggestion
```
We already defined SHOW DESCRIPTION above.  
There is no need to say "must be specified separately" if we clearly define what you CAN do in one statement. Focus on ABILITIES, not INABILITIES.  Only mention inabilities if the feature gap is surprising.

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,104 @@ The scope of the command depends on the user's current scope.
...
+* The wildcard `*` can be used to show descriptions of all queries or all parameters for a specific query.
```
victorleeTG:
```suggestion
* The wildcard `*` can be used to match multiple queries or parameters.
** `*` by itself means all queries in this graph.
**`prefix*` means all queries whose name starts with `prefix`.
** `queryName.*` means all parameters of `queryName`.
```

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,104 @@ The scope of the command depends on the user's current scope.
...
+
+[CAUTION]
+====
+* When updating a query, use `CREATE OR REPLACE QUERY` instead of `CREATE QUERY` to retain the query description.
+* `CREATE QUERY` may unintentionally drop both the query and its description if it already exists, whereas `CREATE OR REPLACE QUERY` ensures the description is retained by replacing the query without dropping it.
+====
+
+=== Show description
+
+The `SHOW DESCRIPTION` command displays the description of a query or a query parameter.
```
victorleeTG:
```suggestion
The `SHOW DESCRIPTION` command displays the descriptions of the queries or query parameters that match the given `objectName`.
```

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,104 @@ The scope of the command depends on the user's current scope.
...
+GSQL > SHOW DESCRIPTION OF QUERY_PARAM query2.*
+----
+
+* For the `SHOW DESCRIPTION` command, you can display the descriptions of individual queries or query parameters. Each query or parameter must be specified separately, and you cannot combine them in a single request.
+* The wildcard `*` can be used to show descriptions of all queries or all parameters for a specific query. 
+* `SHOW DESCRIPTION` also supports a wider range of regular expressions, allowing for more advanced querying when searching for descriptions.
+
+=== Drop description
+
+The `DROP DESCRIPTION` command is used to remove descriptions of queries or query parameters.
```
victorleeTG:
```suggestion
The `DROP DESCRIPTION` removes descriptions of queries or query parameters. 
```

victorleeTG (changes requested):
I gave detailed feedback on SHOW DESCRIPTION.
Please try to understand my feedback and then make appropriate improvements to DROP DESCRIPTION.

Towards the end of the DROP DESCRIPTION command, I think it is appropriate to add a [NOTE] that points out the differences in flexibility of SHOW vs. DROP.  I expect that will have 2 bullet points.

modules/querying/pages/query-operations.adoc
```diff
@@ -1142,3 +1142,111 @@ The scope of the command depends on the user's current scope.
...
+
+Support is provided for dropping descriptions of multiple targets simultaneously.
+You can use the wildcard `*` to remove descriptions of all queries or all parameters of a specific query.
+The query parameter's description is independent of the query's description.
+Therefore, after dropping the query description, the query parameter's description remains unaffected.
+
+[NOTE]
+====
+* `SHOW DESCRIPTION` allows you to display descriptions of multiple queries or query parameters using wildcards like `*` (to show all queries) or `prefix*` (to show queries starting with a specific prefix). You can also use `queryName.*` to show all parameters of a specific query.
+* `DROP DESCRIPTION` supports specifying multiple queries or query parameters in a single command using a comma-separated list. However, it only supports the wildcard `*` for selecting all queries or query parameters and does not support wildcards based on prefixes or individual query names.
```
victorleeTG, line 1251: Please reorganize to **focus** on the differences. I would make the bullets about the features, not about the commands.  E.g.
1. Wildcards:  SHOW ... , but DROP...
2. List of objects:  SHOW ..., but DROP ...

victorleeTG (changes requested):

PR #222, opened 2025-01-22
DOC-2398-Schema Change operation

Tushar-TG-14:


modules/ddl-and-loading/pages/modifying-a-graph-schema.adoc
```diff
@@ -28,6 +28,71 @@ Load or query operations which begin before the schema change will be completed
 Schema changes that are performed during times with high query or data loading volume may have a chance of failure.
 It is considered good practice to perform schema changes during times of low graph activity.
 
+== Schema Change Operations Best Practices
+
+Schema changes are essential for maintaining and improving your graph's structure. However, they must be done carefully to avoid impacting ongoing operations. 
+Following best practices minimizes the risk of failures and ensures smooth execution.
```
victorleeTG: check your grammar

modules/ddl-and-loading/pages/modifying-a-graph-schema.adoc
```diff
@@ -28,6 +28,71 @@ Load or query operations which begin before the schema change will be completed
...
+
+=== Dos and Don'ts During Schema Changes
+
+*Do's:*
+
+* *Do ensure no active queries or loading jobs.* Run schema changes only after aborting ongoing jobs that might reference schema components being modified.
+* *Do perform schema changes during low-traffic periods.* This reduces the chances of failure and conflicts during schema modification.
```
victorleeTG: These seem to be contradictory rules.
Rule 1 says no queries or loading jobs.
Rule 2 says "low traffic". What type of traffic is allowed?  It also does't answer the question of "how much is 'low'?"

modules/ddl-and-loading/pages/modifying-a-graph-schema.adoc
```diff
@@ -28,6 +28,71 @@ Load or query operations which begin before the schema change will be completed
...
+
+Once the schema change is complete, verify data and schema consistency:
+
+*Verifying Data Consistency:*
+
+* Run the `gstatusgraph` command to check that the data size remains consistent with what was recorded before the schema change. This will ensure that no data loss or corruption has occurred during the schema update.
+
+*Verifying Schema Consistency:*
+
+* Use the `gsql` commands to check the schema version after the schema change. This will confirm that the schema updates have been applied correctly, and that no unintended changes have occurred.
```
victorleeTG: What GSQL commands?
@Tushar-TG-14 If you don't know the answer to this question, then the reader won't know either.

modules/ddl-and-loading/pages/modifying-a-graph-schema.adoc
```diff
@@ -28,6 +28,71 @@ Load or query operations which begin before the schema change will be completed
...
+
+=== Impact on Queries and Loading Jobs
+
+Schema changes will invalidate queries and loading jobs dependent on modified schema components.
+This may cause errors or failures in queries trying to reference altered or removed schema elements.
```
victorleeTG: @deepkhajanchi What does "invalidate" mean in this context?
Does it mean "the query and/or loading job" has been "uninstalled" (A query can be uninstalled, but I don't know about loading jobs.)
Or does it mean "the user can still run it, but it might be broken and not work correctly."

victorleeTG: I think we can/should be more specific about the rules for when a query/job will be invalid.

modules/ddl-and-loading/pages/modifying-a-graph-schema.adoc
```diff
@@ -14,8 +14,8 @@ To safely update the graph schema, the user should follow this procedure:
 ====
 A schema change will invalidate any loading jobs or query which relate to an altered part of the schema. Specifically:
 
-* A loading job becomes invalid if it refers to a vertex or and an edge which has been *dropped* (deleted) or *altered* .
-* A query becomes invalid if it refers to a vertex, and edge, or an attribute which has been *dropped* .
+* A loading job becomes invalid if it refers to a vertex or and an edge which has been *dropped* (deleted) or *altered*.
+* A query becomes invalid if it refers to a vertex, and edge, or an attribute which has been *dropped*.
```
victorleeTG: We should always be careful to distinguish between individual vertices/edges vs. vertex/edge **types**.
Loading jobs refer to vertex/edge types; a job will be invalid if it refers to a type or an attribute of a type that no longer exists.

A query usually refers to **types** but in theory could refer to individual vertices and edges.  I think this rule is meant to also refer to types.

victorleeTG: 2nd comment:
Use "DROP" and "DELETE" properly.
In SQL and GSQL, these two keywords have distinct meanings. Therefore, you should never say one is a synonym of the other.
DROP is to remove a catalog item. Catalog items include vertex TYPES, edge TYPES, graphs, queries, and loading jobs. [In a relational database, you could DROP a table.]

DELETE is to remove a data object from the database - a vertex or edge. [In a relational database you could DELETE a row of a table.]

victorleeTG (changes requested):

modules/ddl-and-loading/pages/modifying-a-graph-schema.adoc
```diff
@@ -28,6 +28,71 @@ Load or query operations which begin before the schema change will be completed
...
+
+*Best Practice:* Always abort active queries and loading jobs before performing a schema change.
+
+=== Verifying Schema Integrity After Changes
+
+Once the schema change is complete, verify data and schema consistency:
+
+*Verifying Data Consistency:*
+
+* Run the `gstatusgraph` command to check that the data size remains consistent with what was recorded before the schema change. This will ensure that no data loss or corruption has occurred during the schema update.
```
victorleeTG: Data loss and data corruption are some of the worst errors that a database can do.  Thing about it.
1. Shouldn't we recommend that users perform a backup before performing a schema change?
2. Check other database products to see if/when they mention "data loss", "data corruption", or "data inconsistency".  We want to see how they mention this.  
Compare these two: 
1. "It is essential to make sure that there are no loading jobs in progress when you start the schema change. Failure to follow this rule may result in data corruption." (It's the user's fault.)
2. "Schema changes sometimes result in data corruption." (It's TigerGraph's fault.)

modules/ddl-and-loading/pages/modifying-a-graph-schema.adoc
Tushar-TG-14: @victorleeTG 

This is what Deep has suggested:

From the current version available: [[v4.1: Modifying schema](https://docs.tigergraph.com/gsql-ref/4.1/ddl-and-loading/modifying-a-graph-schema)]

A schema change will invalidate any loading jobs or query which relate to an altered part of the schema. Specifically:
- A loading job becomes invalid if it refers to a vertex or and an edge which has been dropped (deleted) or altered.
- A query becomes invalid if it refers to a vertex, and edge, or an attribute which has been dropped.

Invalid loading jobs are dropped, and invalid queries are uninstalled. After the schema update, the user will need to create and install new load and query jobs based on the new schema.

Please let me know if this answers the question.

modules/ddl-and-loading/pages/modifying-a-graph-schema.adoc
```diff
@@ -28,6 +28,58 @@ Load or query operations which begin before the schema change will be completed
...
+ 
+* A loading job becomes invalid if it references a vertex or edge that has been DROPPED or ALTERED. 
+* A query becomes invalid if it refers to a DROPPED vertex, edge, or attribute.
+
+Invalid loading jobs are *DROPPED*, and invalid queries are *uninstalled*. After the schema update, users must create and install new loading and query jobs based on the updated schema.
+
+*What to Do:*
+
+* After schema changes, reinstall and redefine queries and loading jobs based on the new schema.
+* Ensure that the updated queries and jobs align with the modified schema and are properly tested to ensure they work as expected.
```
victorleeTG:
```suggestion
```
This part is (1) repeating what you said earlier and (2) applies to all schema changes, not just DSC.
That is, a query or loading job does not have to be running in order for it to be invalidated.
So, I would remove this.  But you can see whether you think your wording was better here or above and use the one with better wording.

victorleeTG (changes requested):

PR #227, opened 2025-02-04
DOC-2413-GSQL-OpenCypher-behaves-differently-from-Neo4j-because-we-are-schema-first-4.1.2

anjaliIthapeTG:


modules/openCypher-in-gsql/pages/openCypher-in-gsql.adoc
```diff
@@ -323,3 +323,40 @@ WHERE x.friendCount = y.friendCount
...
+Neo4j does not enforce a schema, so the query:
+
+* Processes all vertex types, regardless of their attributes
+* Returns values for existing attributes but returns `NULL` for attributes that do not exist
+
+==== Importance
+
+This difference is crucial for OpenCypher users migrating from Neo4j to GSQL. In GSQL, queries must align with the schema, meaning attributes must be explicitly defined in the vertex type. In contrast, Neo4j allows queries to access attributes flexibly, even if some results contain NULL values.
+
+By understanding this distinction, users can write more effective OpenCypher queries tailored to GSQL’s schema-first approach.
```
DavidFan2000, line 359: I'm not sure if this line is necessary.

modules/openCypher-in-gsql/pages/openCypher-in-gsql.adoc
```diff
@@ -323,3 +323,40 @@ WHERE x.friendCount = y.friendCount
...
+==== How Neo4j Interprets This Query
+
+Neo4j does not enforce a schema, so the query:
+
+* Processes all vertex types, regardless of their attributes
+* Returns values for existing attributes but returns `NULL` for attributes that do not exist
+
+==== Importance
+
+This difference is crucial for OpenCypher users migrating from Neo4j to GSQL. In GSQL, queries must align with the schema, meaning attributes must be explicitly defined in the vertex type. In contrast, Neo4j allows queries to access attributes flexibly, even if some results contain NULL values.
```
DavidFan2000: Good point! This difference is crucial for users who are migrating. 

I think the main implication is that when a query tries to access non-existing attributes, the difference can either lead to different return values or reject the query for no matching vertex types. Users should be aware of this when they design their schema or adapt their queries.

anjaliIthapeTG: Added above point in the documentation

PR #232, opened 2025-02-14
DOC-2413-GSQL-OpenCypher-behaves-differently-from-Neo4j-because-we-are-schema-first-4.1.2

anjaliIthapeTG:


DavidFan2000 (approved):

modules/openCypher-in-gsql/pages/openCypher-in-gsql.adoc
```diff
@@ -354,12 +354,12 @@ OpenCypher does not enforce a schema, so the query:
 
 ==== Importance
```
victorleeTG: This is an unusual heading name.
And I don't think it should be a subsection of the example.  Because these are the "rules" for how OpenCypher in GSQL behaves, it would either be at the same or higher level than then example.

modules/openCypher-in-gsql/pages/openCypher-in-gsql.adoc
```diff
@@ -324,9 +324,9 @@ WHERE x.friendCount = y.friendCount
 
 * Pattern fragments (x:user) and (y:user) *are not* connected by edge traversal or by sharing vertex variables.
 
-== Key Differences: GSQL vs. OpenCypher
+== Key Differences: GSQL OpenCypher mode vs. OpenCypher
 
-GSQL is strongly typed and **schema-first**, while OpenCypher is **schema-optional**. This fundamental difference impacts how OpenCypher queries behave in each system.
+GSQL OpenCypher is strongly typed and **schema-first**, while OpenCypher is **schema-optional**. This fundamental difference impacts how OpenCypher queries behave in each system.
```
victorleeTG: "GSQL OpenCypher" sounds strange to me because GSQL is a graph query language (and more) , and OpenCypher is another graph query language.  The direct joining of the two words doesn't make semantic sense. It's like saying "Toyota Honda car". Previously on this page, we used the more explicit phrases "OpenCypher in GSQL" and "GSQL OpenCypher mode." For the later, it would be even more appropriate to say "GSQL's OpenCypher mode" (use the possession form of GSQL.)

When you want to define a new, briefer terminology, the standard method is to give the full name, then the abbreviate term in parentheses and quotation marks, such as

GSQL's OpenCypher mode ("GSQL OpenCypher") is strongly type.

modules/openCypher-in-gsql/pages/openCypher-in-gsql.adoc
```diff
@@ -324,9 +324,9 @@ WHERE x.friendCount = y.friendCount
...
 
 === Example: OpenCypher Query Behavior
```
victorleeTG, line 331: The example really should have some sample data and the output from GSQL's OpenCypher vs. standard OpenCypher. The verbal description alone may not be that clear.

victorleeTG (changes requested):
The example does not have any actual

victorleeTG:
replaced by an updated PR

victorleeTG (approved):

PR #237, opened 2025-02-25
DOC-2496-openCypher syntax

Tushar-TG-14:


modules/openCypher-in-gsql/pages/openCypher-in-gsql.adoc
```diff
@@ -7,6 +7,11 @@ TigerGraph's GSQL query language supports many openCypher features. This page li
 
 To see how to run openCypher on the TigerGraph platform, see xref:openCypher-in-gsql.adoc[].
 
+[NOTE]
+====
+From version *4.1.2*, certain openCypher syntaxes are now fully supported, eliminating previous restrictions.
```
victorleeTG:
```suggestion
From version 4.1.2, certain restrictions on the `MATCH` and `WITH` clauses have been eliminated.
```
1. "syntax" already refers to a collection (multiple) rules, so plural "syntaxes" is not really correct here.
2. I think we still have one restriction on MATCH, so unfortunately we cannot say fully supported.

victorleeTG (changes requested):

PR #242, opened 2025-02-28
Doc-2526-New-operations-with-bitwise-accum-[4.2.0-dev]

srinadhmamidalaTG:


modules/querying/pages/accumulators.adoc
```diff
@@ -672,6 +672,9 @@ Overflow on the left is ignored.
 * When a bitwise accumulator length is longer than 64 bit, it's assigned using an array of two integers.
```
victorleeTG:
```suggestion
* When a bitwise accumulator length is longer than 64 bits, it's assigned using an array of integers, one integer for each group of 64 bits, e.g., [int3, int2, int1].
```
@srinadhmamidalaTG We need to fix the doc error: it shouldn't say "two". 
Also, note that I numbered the integers 3, 2, 1, not 1, 2, 3.  That is to be conceptually consistent with the next sentence.

modules/querying/pages/accumulators.adoc
```diff
@@ -672,6 +672,9 @@ Overflow on the left is ignored.
...
 Each integer is converted to a 64-bit sequence of 1s and 0s.
 The integer in the second position will take up the first 64 bits from the right of the sequence, and the integer on the left will take up the remaining bits.
```
victorleeTG:
```suggestion
The rightmost (last) element in the array is for the 64 least significant bits. If the total number of bits is not a multiple of 64, then the leftmost (first) element in the array will be for fewer than 64 bits.
```
"second position" is not clear. A clearer way to talk about the ends of an array are start/end, leftmost/rightmost.  For the digits/bits in a number,  "least significant" and "most significant" are precise terms.

modules/querying/pages/accumulators.adoc
```diff
@@ -672,6 +672,9 @@ Overflow on the left is ignored.
...
+* The previous documentation incorrectly suggested a maximum length of 2 x 64 bits (128 bits) for bitwise accumulators. There is no length limit in the latest version, though bugs affecting functions longer than 64 bits have been fixed in version 4.2.0.
+* Long bitwise accumulators can be initialized using a list of 64-bit integer values:[value1, value2, value3, etc.]
```
victorleeTG:
```suggestion
```
We no longer need this, because we already explained it above.

modules/querying/pages/accumulators.adoc
```diff
@@ -672,6 +672,9 @@ Overflow on the left is ignored.
 * When a bitwise accumulator length is longer than 64 bit, it's assigned using an array of two integers.
 Each integer is converted to a 64-bit sequence of 1s and 0s.
 The integer in the second position will take up the first 64 bits from the right of the sequence, and the integer on the left will take up the remaining bits.
+* The previous documentation incorrectly suggested a maximum length of 2 x 64 bits (128 bits) for bitwise accumulators. There is no length limit in the latest version, though bugs affecting functions longer than 64 bits have been fixed in version 4.2.0.
```
victorleeTG:
```suggestion
```
We don't describe past bugs in the main documentation.  
1. If we have a known bug that has not been been fixed, we put a brief warning in the relevant section and in the release notes, so that the user will know how to avoid the bug. When the bug is fixed, we remove the warning.
2. In this case, I think the bug only occurred when the developer (and TigerGraph's graph algorithm team) tried to use mathematical functions which weren't officially supported. Therefore, we don't have to describe the bug at all.

modules/querying/pages/accumulators.adoc
```diff
@@ -772,6 +775,28 @@ Returns a bitwise accumulator whose every bit is the `AND` result between the tw
 |Mutator
 |Compares two bitwise accumulators of the same length.
 Returns a Bitwise accumulator whose every bit is the `OR` result between the two bitwise accumulators.
+
+| `.& ( accum )`
+|
+|
+|Compares two bitwise accumulators of the same length and returns a new bitwise accumulator whose every bit is the `AND` result of the corresponding bits in the current bitwise accumulator and the given one, without modifying the current accumulator.
```
victorleeTG, line 780: For added clarity, you might want to give a more specific name to "the given one", e.g.,
`.& (arg_accum)`
then you can say "Compares this accumulator with the given accumulator `arg_accum` and returns..."

Also, I replaced "current" with "this" because in the C++ and Java languages, "this" has the special meaning of referring to the current object.

modules/querying/pages/accumulators.adoc
```diff
@@ -1424,6 +1449,17 @@ Comparison (and assignment with the `+=` operator) is only possible if two HeapA
...
+
+**Example:** WHERE (@@BitwiseAccum1 > 32) AND (@@BitwiseAccum2 < 1024)
+
+**Implementation Differences**
+
+**Standard Bitwise Accum:** Implicitly convert it to an integer for comparison.
+
+**User-defined Bitwise Accum:** Comparators are overridden for bitwise accumulators longer than 64 bits.
+
+
```
victorleeTG: This section confused me at first, and there are a number of issues with how it is written:
1. It is in the HeapAccum section but it is talking about Bitwise Accumulators.
2. Comparator support: This sentence seems very terse to describe important new functionality. 
How do other accumulators talk about their support for comparators?
Note, we have a GSQL section that describes Comparison Operators: https://docs.tigergraph.com/gsql-ref/4.1/querying/operators-and-expressions#_comparison_operators_and_conditions
Maybe the answer is to say that bitwise accumulators support comparison operators, with a link to that section.
3. Be careful about saying "including" because that word implies you are listing SOME BUT NOT ALL valid cases. I don't think that is the meaning you want to convey here.
4. In main documentation pages, we avoid saying "now" (and very strongly discourage "new") because you (the writer) are referring to your own time reference which is not necessarily the reader's time reference. Unless there is a reason to modify it again, this page will be copied from v4.2 to v4.3, to v.4.4, etc. Then "new" will be clearly wrong. If there is a reason to point out the difference in functionality between versions, name a specific number as the boundary, e.g., "Beginning with v 4.2.0, Bitwise accumulators support comparator operators".
5. The user documentation for bitwise accumulator does not use the terms "standard" and "user-defined". We generally don't discuss implementation issues with users, unless it will help them to understand behavior.  If I remember correctly what Dadong explained, I would boiled it down to
"NOTE: Bitwise accumulators longer than 64 bits will perform comparison operators slower than default length accumulators." ("Default length" is the correct concept, as described in the current documentation.)
6. Don't abbreviate "accumulator" to "accum". Some other abbreviates we use in our informal chat but which should never be used in documentation:

TG for TigerGraph
config for configuration (noun) or configure (verb). However, `config` is a keyword in the language. Use it only when taking about that command.

victorleeTG (changes requested):

victorleeTG:
replaced with a newer PR

PR #246, opened 2025-03-03
Doc-2455-support-datetime-parameter-type-when-running-interpret-query-through-endpoint-api [4.2.0-dev]

srinadhmamidalaTG:


modules/querying/pages/func/datetime-functions.adoc
```diff
@@ -136,7 +136,7 @@ The default value for this parameter is `+"%Y-%m-%d %H:%M:%S"+`
 
 [source,text]
 ----
-datetime_format(to_datetime("2020-01-02 05:30:12"), "hi, it's %Y-%m-%d")
+(datetime_format("2020-01-02 05:30:12"), "hi, it's %Y-%m-%d")
```
victorleeTG, line 139: This example is fine, but it would be a clearer and more powerful example to show how datetime_format enables you to invoke individual parts of a datetime, such as 
"hi, it's %d/%m in the year %y.") -> "hi, it's 01/02 in the year 2020."

victorleeTG (approved):

PR #252, opened 2025-03-06
DOC-2008-Added a section on `=` operator [3.9]

Tushar-TG-14:


modules/querying/pages/accumulators.adoc
```diff
@@ -1001,6 +1001,27 @@ The behavior is identical.
 
 The two MapAccums must have identical data types.
 
+The `=` operator in MapAccum is used to *overwrite* the value of an existing key in the collection. 
+When you use the `=` operator, the value for the specified key is replaced with the new value, discarding any previous accumulated values for that key.
```
victorleeTG: Sorry, you missed the point.

The `=` operator in MapAccum is used to *overwrite* the value of the entire map.
All existing key-value pairs are erased. The MapAccum is restarted with the key-value on the right side of the `=` operator.

victorleeTG (changes requested):

modules/querying/pages/accumulators.adoc
victorleeTG: Your interpretation was what the customer wanted and what our solution engineers incorrectly thought was our behavior. Our design engineers didn't understand the issue well enough and said yes, that particular key-value is overwritten, but they didn't mention that ALL OF THE REST OF THE MAPACCUM IS ERASED.

victorleeTG (approved):

PR #256, opened 2025-03-11
DOC-2562 Doc Change for GSQL endpoint: Add sample[4.1]

srinadhmamidalaTG:


modules/API/pages/gsql-endpoints.adoc
```diff
@@ -3076,6 +3076,153 @@ Sample Response::
...
+----
+CREATE QUERY q1(int p1){print p1;}
+----
+--
+Sample Request::
++
+--
+[source,bash]
+----
+curl -s -u tigergraph:tigergraph -X POST "localhost:14240/gsql/v1/queries/q1?graph=ldbc_snb" -H 'Content-Type: application/json' -d '{"diagnose":false,"denseMode":false,"allvertex":false,"asyncRun":false,"parameters":{p1:1}}'
```
victorleeTG:
```suggestion
curl -H 'Content-Type: application/json' -d '{"parameters":{p1:1}}' -X POST "localhost:14240/gsql/v1/queries/q1?graph=ldbc_snb" 

```
1. I had asked you to remove `-s -u tigergraph:tigergraph`.  I expect you to remember this. 
2. We don't need `"diagnose":false,"denseMode":false,"allvertex":false,"asyncRun":false`
They are optional, and we are setting them to false, which is already the default. Total waste.
3. Move the `-d` payload to the beginning to make it easier for the reader to see what we want them to focus on: the parameter.

modules/API/pages/gsql-endpoints.adoc
```diff
@@ -3076,6 +3076,153 @@ Sample Response::
 
 '''
 
+==== Query Parameter Types
+
+This section provides an overview of different parameter types used in GSQL queries, including their usage, sample queries, request formats, and responses.
+
+=== Primitive type Parameter
```
victorleeTG, line 3083: Heading level is wrong.  If you change the level of the header above (Query Parameter Types, from 2 to 4), then all the subheadings have to change too (e.g., from 3 to 5).

However, you have TOO MANY LEVELS. Never use level 5. If you have level 5, rethink your document. You don't need to make everything a header. I would just make this bold text, not a header.

modules/API/pages/gsql-endpoints.adoc
```diff
@@ -3076,6 +3076,153 @@ Sample Response::
...
 '''
 
+==== Query Parameter Types
+
+This section provides an overview of different parameter types used in GSQL queries, including their usage, sample queries, request formats, and responses.
+
+=== Primitive type Parameter
+
+
+==== Example
```
victorleeTG, line 3086: Another unnecessary heading. What you should have instead is a sentence to tell user what you want them to notice the the example below. Just showing a block of code is not user-friendly. You have to call out what you want them to notice.

victorleeTG (changes requested):

PR #262, opened 2025-03-24
DOC-2415-Added a page FROM Clause (v2)

Tushar-TG-14:


modules/querying/pages/select-statement/FROM Clause (v2)
```diff
@@ -0,0 +1,372 @@
+= FROM Clause (v2)
```
victorleeTG:
```suggestion
= FROM Clause (Syntax V2)
:navtitle: From Clause (V2)

```
@Tushar-TG-14 Note a couple of details:
1. In the body of the document, V1 and V2 are capitalized, so it should be capitalized in the title also. 
2. I decided it's better to have "Syntax V2" in the title to be clear we are talking about GSQL Syntax, not about the database version.
3. BUT, I still want the navigation title to be short. There is a special directive for setting the navigation title differently than the page title.
4. There should be a blank line between the title and the body.

victorleeTG (approved):
@Tushar-TG-14 The page contents are fine (thought I tweaked the title section), but the page/file name is not.  I'm going to accept this, but you should rename the page.

"FROM Clause (v2)" - NEVER have blank spaces in the filenames. This is not Windows or MacOS.  While modern Linux systems can handle blank spaces, it creates some ugly things in places.

Use this name: "from-clause-v2.adoc"

PR #274, opened 2025-04-02
DOC-2487-Added a new page for Query Plan Caching

Tushar-TG-14:


modules/querying/pages/query-optimizer/Query Plan Cache
```diff
@@ -0,0 +1,54 @@
+= Query Plan Cache
+
+This page covers how TigerGraph handles query execution and optimization through query plan caching. 
+When a query is executed, the system parses, transforms, and optimizes it before generating a query plan in JSON format to guide its execution. 
+While these steps are essential for processing the query, they can introduce significant overhead, especially when the same query is executed multiple times. 
+To mitigate this, TigerGraph caches the generated query plan in memory, ensuring faster execution on subsequent queries.
```
victorleeTG:
```suggestion
To mitigate this, TigerGraph caches the generated query plan in memory, ensuring faster execution on subsequent executions of a query.
```

modules/querying/pages/query-optimizer/Query Plan Cache
```diff
@@ -0,0 +1,54 @@
...
+
+== Query Normalization
+
+In many cases, users may frequently change constants in their queries. These changes can invalidate the cached query plan, making it necessary to re-interpret the query. 
+To handle this issue, TigerGraph employs a process known as *query normalization*.
```
victorleeTG:
```suggestion
In many cases, users may frequently change constants in their queries. These changes would invalidate the cached query plan, making it necessary to re-interpret the query. 
To avoid the reinterpretation, TigerGraph employs a process known as *query normalization*.
```

PR #278, opened 2025-04-03
DOC-2449-Added Query Profiling Page

Tushar-TG-14:


modules/querying/pages/query-optimizer/query-profiling.adoc
```diff
@@ -1,2 +1,426 @@
...
+** `/allpaths`
+** `/searchvertex`
+
+For these queries, profiling data will include more detailed information about memory usage and the overall query execution, as explained in the xref:{page-component-version}@gsql-ref:querying:query-optimizer/query-profiling.adoc[Profiling Data Explanation] section below.
+
+== Prerequisite
+
+To enable query profiling, users must have the READ QUERY or OWNERSHIP privileges for the query being executed. 
+If RESTPP authentication is enabled, users will need to ensure they have the appropriate privileges. 
+Otherwise, they may encounter an error message like this:
```
victorleeTG, line 37: I think this error message is not unique to this feature.  We may want to consider moving this to the RESTPP authentication page instead.

modules/querying/pages/query-optimizer/query-profiling.adoc
```diff
@@ -1,2 +1,426 @@
 = Query Profiling
-:description: Query Profiling provides data to analyze the details of a query execution.
\ No newline at end of file
+:toclevels: 1
+:description: Query Profiling provides data to analyze the details of a query execution.
+
+This page provides an overview of query profiling in TigerGraph, a feature introduced in version 4.2.0.
```
victorleeTG:
```suggestion
This page provides an overview of query profiling in TigerGraph, a feature introduced in version 4.2. 
```
If a feature is introduced in the initial Feature Release (Preview), we should not mention the ".0". We only want to use the 3rd number when we have to.

modules/querying/pages/query-optimizer/query-profiling.adoc
```diff
@@ -1,2 +1,426 @@
...
+      }
+    }
+  },
+  "code": "REST-15001"
+}
+----
+
+== How to Activate Profiling
+
+In TigerGraph version 4.2.0, profiling can be activated using the *PROFILE* flag. Currently, only one profiling mode, `BASIC`, is available.
```
victorleeTG:
```suggestion
Activate profiling by setting a profiling mode. Currently, only one profiling mode, `BASIC`, is available.
```
I'm thinking about whether "activate" is the best word, or "enable".  It is chosen for each individual query when you run it. I.e., it does not "stay" on for a 2nd query.

victorleeTG (approved):
Still needs some improvement around the Run Options, but accepting this as a draft.

PR #300, opened 2025-08-13
DOC-2946: Documented known limitations and issues

Tushar-TG-14:


modules/vector/pages/index.adoc
```diff
@@ -142,6 +142,8 @@ CREATE GLOBAL SCHEMA_CHANGE JOB add_emb2 {
 RUN GLOBAL SCHEMA_CHANGE JOB add_emb2
 ----
 
+When modifying vector attributes with a schema change job and using the `-N` option, *all queries become invalid* and must be reinstalled, even if they are unrelated to the change.
```
victorleeTG: According to the documentation for RUN SCHEMA CHANGE (https://docs.tigergraph.com/gsql-ref/4.2/ddl-and-loading/modifying-a-graph-schema#_run_schema_change_job), invaliding queries is the primary purpose of -N.  (I wonder if that is really true?  Is there another purpose of -N that the developers forgot to tell us about?)
Regardless, I don't see why we need a special warning on the vector page.  -N is not default. If you choose to use it, it clearly says it will invalidate queries.

modules/vector/pages/index.adoc
```diff
@@ -186,6 +191,17 @@ The search index is used when the `vectorSearch` function is invoked.
...
+During export, TigerGraph does not wait for the vector index rebuild to finish. If you export before rebuild completes, some vectors may be missing.
+
+To check rebuild status:
+[source.wrap,bash]
+----
+curl -s -X GET "http://<host>:14240/restpp/vector/status"
+----
+
+* Rebuild_processing means the index is still rebuilding.
+* Ready_for_query means rebuild is complete.
```
victorleeTG:
```suggestion
* `Rebuild_processing` means the index is still rebuilding.
* `Ready_for_query` means rebuild is complete.
```

modules/vector/pages/index.adoc
```diff
@@ -186,6 +191,17 @@ The search index is used when the `vectorSearch` function is invoked.
 The index is automatically updated in respond to changes in the vector database.
 Because similar calculations take some time, the index can lag behind the data updates.
 
+If you have many vertex types, index rebuild time increases proportionally, even if the number of vector records is the same.
+During export, TigerGraph does not wait for the vector index rebuild to finish. If you export before rebuild completes, some vectors may be missing.
```
victorleeTG:
```suggestion
If you have many vertex types, index rebuild time increases proportionally, even if the total number of vector records is the same.

[WARNING]
====
Vectors which have not yet been indexed will not be included in vector search results.
A database export or backup may not include vectors which have not yet been indexed.
====

```

modules/vector/pages/index.adoc
```diff
@@ -226,6 +242,11 @@ E.g., `{candidate_set: vset1, ef: 20, distance_map: @@distmap}`.
 
 The meaning of the optional parameters is best explained by example.  See the link:https://github.com/tigergraph/ecosys/blob/master/tutorials/VectorSearch.md[Vector Search tutorial].
 
+In version 4.2.1, there are some limitations when using vectorSearch() in GSQL queries:
```
victorleeTG: These are not 4.2.1-specific limitations. These are simply THE limitations of vector search.  Vector search was introduced in 4.2.0, so there is not yet any evolution or improvement. This is just the way the feature works, for now.

*Limitations* should be a section.

modules/vector/pages/index.adoc
victorleeTG: The developers didn't mention Backup, but I wonder if it has the same limitation as Export.
I would like to understand the problem a little better.

victorleeTG (changes requested):

PR #308, opened 2025-10-15
DOC-2200 supporting null val

AyeshaFirdausTG:


modules/querying/pages/data-modification-statements.adoc
```diff
@@ -340,6 +340,23 @@ VALUES (value_for_from_vertex, value_for_to_vertex, <1>
 For each attribute value, provide either an expression _expr_ or `_`, which means the default value for that attribute type.
 The optional _name_ which follows the first two (id) values is to specify the source vertex type and target vertex type, if the edge type had been defined with wildcard vertex types.
 
++
+.NULL Value Ingestion
++
+The `INSERT INTO` statement supports inserting `NULL` values for attributes defined as `NULLABLE` in the schema.
+*  `NULL` or `null` can be used in the `VALUES` clause to insert a NULL value.
```
victorleeTG, line 347:
```suggestion
*  Either `NULL` or `null` can be used in the `VALUES` clause to represent a NULL value.
```

modules/querying/pages/data-modification-statements.adoc
```diff
@@ -340,6 +340,23 @@ VALUES (value_for_from_vertex, value_for_to_vertex, <1>
 For each attribute value, provide either an expression _expr_ or `_`, which means the default value for that attribute type.
 The optional _name_ which follows the first two (id) values is to specify the source vertex type and target vertex type, if the edge type had been defined with wildcard vertex types.
 
++
+.NULL Value Ingestion
```
victorleeTG, line 344: This section does not seem to be in the right place on this page.
You have it as a continuation of the section on " two options for specifying the attributes of the vertex or edge type for the values provided:"  But NULLABLE is about the command in general, not about these two options.

I have more to say on this, which I will write in a separate note.

modules/ddl-and-loading/pages/modifying-a-graph-schema.adoc
```diff
@@ -167,7 +167,7 @@ ADD VERTEX::
 --
 ----
 ADD VERTEX Vertex_Type_Name ( PRIMARY_ID id_name id_type
-    [, attribute_name type [DEFAULT default_value] ]* )
+    [, attribute_name type  [NULLABLE][DEFAULT default_value] ]* )
```
victorleeTG, line 170: This syntax detail is correct, but there is another section on the page that needs to same treatment.
Besides the A**DD VERTEX/EDGE section**, there is an analogous
**ADD VERTEX/EDGE (global)** section.
https://docs.tigergraph.com/gsql-ref/4.2/ddl-and-loading/modifying-a-graph-schema#_add_vertex_edge_global

modules/ddl-and-loading/pages/creating-a-loading-job.adoc
```diff
@@ -108,6 +108,72 @@ RUN LOADING JOB job1 USING file1="m1:/data/v1_1.csv", file2="m2:/data/e2.csv" <2
...
+----
+"results": [
+  { "1,null": null },
+  { "2,NULL": null },
+  { "3,": null },
+  { "4,Null": "Null" }
+]
+----
+
+==== Loading NULL during creation of loading job
```
victorleeTG, line 150: I would modify and reduce the scope of this section.
There are basically 3 ways to end up with NULL in the loaded data:
1. Because NULL (including empty fields) was in the input data.
>That is already covered in the section above.

2. Because the loading job has an explicit value (NULL) instead of making a reference (e.g. $1) to a source file.
>This can happen with any data type (explicit INT, explicit BOOLEAN, ...), so there does not seem to be a reason to make an exception for NULL.

3. Because the loading job uses an underscore "_" to skip an attribute.
>This is discussed but not that clearly, and NULL changes the odd behavior, so we should write about this.

We might want a section called "Omitted Fields and Default Values".
Here we would:
* Talk about omitted fields in CSV, JSON, and loading job "destination clauses": if they are legal and how they are interpreted. If legal, then the default value is used.
* Explain the rules for default values.
** Depends if the particular vertex or edge type defined a custom default value for that attribute
** Depends on whether the attribute is NULLABLE or not

modules/ddl-and-loading/pages/attribute-data-types.adoc
```diff
@@ -120,6 +120,10 @@ Existing schemas that are using `STRING COMPRESS` can continue to function norma
 If you need reference for `STRING COMPRESS`, please refer to GSQL Language Reference version 3.5 or older.
 ====
 
+[NOTE]
+====
+Primitive data types used as attributes of a vertex or edge, can be set to `NULL`.
+====
```
victorleeTG, line 126: I would make this a new (brief) section on Null Values, not a Note.
* If an attribute is defined as NULLABLE, then its value can be NULL, in addition to its characteristic values
* explain what it means if an attribute is null
* explain how to format null in as an input

modules/ddl-and-loading/pages/attribute-data-types.adoc
```diff
@@ -143,6 +147,17 @@ Due to multithreaded GSQL loading, the initial order of elements loaded into a l
...
+* `LIST`, `SET`, `MAP` can be assigned `NULL` as an attribute.
+* For `LIST` or `SET`:
+** `NULL` is different from [NULL, NULL] or {NULL}.
+** Elements inside `LIST` or `SET` cannot be `NULL`.
+* For `MAP`:
+** Keys and Values cannot be `NULL`.
+
+====
+
```
victorleeTG, line 160: Not all of this belongs in a special note:
> * `LIST`, `SET`, `MAP` can be assigned `NULL` as an attribute

This is just following the standard rule that any attribute that is NULLABLE can have null.  No special treatment.

> ** Elements inside `LIST` or `SET` cannot be `NULL`. 

This does deserve being in a NOTE because this is where TigerGraph's support is not as complete as it is in some other products.

> * For `MAP`:
> ** Keys and Values cannot be `NULL`

I would just lump this together with `LIST` and `SET`.

> ** `NULL` is different from [NULL, NULL] or {NULL}. 

From a technical standpoint, this is not saying anything that isn't already said in the preceding rules. It's usefulness is as an example, not an additional rule.

MyList = NULL   // This is setting the entire list to null. Supported.
MyList = [NULL] // This is creating a list with one element whose value is null. Not supported.

modules/ddl-and-loading/pages/creating-a-loading-job.adoc
```diff
@@ -108,6 +108,72 @@ RUN LOADING JOB job1 USING file1="m1:/data/v1_1.csv", file2="m2:/data/e2.csv" <2
 <1> File path specified at compile time.
 <2> Run-time specification will override path specified at compile time.
 
+=== Loading NULL Values
```
victorleeTG, line 111: We already have sections that talk about how to load each type of data. It is called "Attributes and attribute expressions", with a subsection for each data type.
I would move this section there, and follow the format used there.

victorleeTG (changes requested):

PR #317, opened 2026-01-16
Update output-statements-and-file-objects.adoc

AyeshaFirdausTG:


modules/querying/pages/output-statements-and-file-objects.adoc
```diff
@@ -12,9 +12,8 @@ A `PRINT` statement can appear wherever query-body statements are permitted.
 A `PRINT` statement does not trigger immediate output.
 The full set of data from all PRINT statements is delivered at one time, when the query concludes.
 
-A query can print a maximum of 2GB of data.
-
-If the output is to a `FILE` object, then the size limit does not apply.
+If the query response is large, it is recommended to use `PRINT … TO_CSV`.
```
pingxieTG, line 15: "Large" can be too vague; it can be several gigabytes.

pingxieTG: We may give a redirection link to the "`PRINT … TO_CSV ...`"

PR #324, opened 2026-03-18
DOC-2101-data-modification-statements-3.10

swatiMavinakattiTG:


Tushar-TG-14:
@swatiMavinakattiTG This looks good and addresses the gap for set and map examples.

I have a couple of small suggestions before we finalize:

1. Add a short explanation before the example to clarify the syntax differences ([] for list, () for set, -> for map). This will make it easier for users to understand.
2. Make the attribute comments consistent (e.g., “list-type”, “set-type”, “map-type”) and optionally mention syntax usage.
3. Minor grammar fix: “PRINT S still show” → “PRINT S still shows”.

PR #333, opened 2026-05-29
Update creating-a-loading-job.adoc

priyankagunaki-cloud:


modules/ddl-and-loading/pages/creating-a-loading-job.adoc
```diff
@@ -338,7 +338,7 @@ The loader accepts any of the following string formats:
 * `%Y-%m-%dT%H:%M:%S.000z` (e.g., 2011-02-03T01:02:03.123z, 123 will be ignored)
 * `%Y-%m-%d` (only date, no time, e.g., 2011-02-03)
 * `%Y/%m/%d` (only date, no time, e.g., 2011/02/03)
-* Any integer value (Unix Epoch time, where Jan 1, 1970 at 00:00:00 is integer 0)
+* Any string value (Unix Epoch time, where Jan 1, 1970 at 00:00:00 is 0)
```
Tushar-TG-14: @priyankagunaki-cloud Since only Unix Epoch timestamps represented as strings are valid, can we make the wording more explicit? Something like:
A Unix Epoch timestamp represented as a string (for example, "0" represents Jan 1, 1970 at 00:00:00 UTC)

This would make the expected format clearer to users.

modules/ddl-and-loading/pages/creating-a-loading-job.adoc
Tushar-TG-14: The format notation says %m and %s are a month and a day, respectively. Should %s be %d here? It looks like a typo in the date format description.
Update it to:
%m and %d are a month (1 to 12) and a day (1 to 31), respectively. Leading zeroes are optional.

Tushar-TG-14: Fix the grammar:
When loading data into a DATETIME attribute, the GSQL loader automatically reads a string representation of datetime information and converts it to an internal DATETIME representation.

Tushar-TG-14 (changes requested):

PR #346, opened 2026-06-23
DOC-1741 Update data-types.adoc

priyankagunaki-cloud:


modules/querying/pages/data-types.adoc
```diff
@@ -220,6 +220,27 @@ a| A decimal precise up to about 7 digits: `3.14159`
 | No literal. Can be converted from a correctly formatted string with xref:querying:func/type-conversion-functions.adoc#_parse_json_array[`parse_json_array()`].
 |===
 
+[CAUTION]
+====
+When passing a `DATETIME` value as a query parameter via a REST API call,
+do not enclose the value in single or double quotes.
+Quoted values will not return an error but will silently return `0` results.
```
Tushar-TG-14: Replace this sentence with:
Quoted values are not parsed as valid DATETIME parameters. The query may execute successfully, but DATETIME-based comparisons can return incorrect results, such as returning no matching records.

Tushar-TG-14 (changes requested):

modules/querying/pages/data-types.adoc
priyankagunaki-cloud: Okay, Sure.

PR #351, opened 2026-07-08
DOC-1227 Add Desc user_defined_header

priyankagunaki-cloud:


modules/ddl-and-loading/partials/load-statement.adoc
```diff
@@ -286,8 +286,7 @@ Both values are strings and must be enclosed in double quotes.
 Default is `"false"`
 
 |`USER_DEFINED_HEADER`
-|Specifies the name of the header variable, when a
-header has been defined in the loading job, rather than in the data file
+|Specifies the name of a header variable defined by a preceding DEFINE HEADER statement. The loader uses the user-defined column names instead of the column names in the input file. For an example, see xref:data-loading:load-local-files.adoc#user-defined-header-example[Example: Using USER_DEFINED_HEADER].
```
victorleeTG:
```suggestion
|Instructs the loading job to use the user-defined column names defined by a preceding DEFINE HEADER statement instead of the column names in the input file or the generic positional indexes ($0, $1, etc.). For an example, see xref:{page-component-version}@tigergraph-server:data-loading:load-local-files.adoc#user-defined-header-example[Example: Using USER_DEFINED_HEADER].
```
Because you are referencing a different Antora component (gsql-docs to server-docs), the xref path needs to specify the component (and in most cases you also want to tell it to go the same component version as the source version {page-component-version}@)

victorleeTG (changes requested):

priyankagunaki-cloud:
@victorleeTG I have addressed your comment. Please check.

Tushar-TG-14 (approved):

PR #352, opened 2026-07-14
DOC-1638 Add multiple edge section

priyankagunaki-cloud:


modules/ddl-and-loading/pages/defining-a-graph-schema.adoc
```diff
@@ -373,6 +373,16 @@ For example:
 CREATE DIRECTED EDGE Member_Of (FROM Person, TO Org | FROM Org, TO Org, joined DATETIME)
 ----
 
+=== Multi-edges
```
victorleeTG: @priyankagunaki-cloud The ticket said to add a section, so you did. Now consider the resulting page.
You have a new Heading 3 section about Multi-edges, which describes why they are useful and then says the way to create them is with a Discriminator. There is an example, which was copied from the Discriminator section. This is immediately followed by a Heading 3 section (same level) which describes  Discriminators. It again shows the same example.

Does this not seem redundant to you?

This is why I say do not blindly follow tickets. A ticket is trying to describe a problem or need. Understand the need, then figure out the best way to address it. The person who wrote the ticket may not have put in the time to understand the full context, or they may not be the best writer or explainer.

victorleeTG (changes requested):

modules/ddl-and-loading/pages/defining-a-graph-schema.adoc
priyankagunaki-cloud: @victorleeTG 
Updated the PR based on the review feedback by removing the redundant Multi-edges section and instead introducing the concept of multi-edges in the CREATE EDGE introduction. The implementation details remain in the DISCRIMINATOR section, with a cross-reference added for users to learn more.
Please review.

victorleeTG (approved):

